PR #241, opened 2026-02-27
fix: make ChatDatabase resilient to SQLite connection failures

StephaneDelcroix:
## Problem

`ChatDatabase` methods threw unobserved task exceptions when the SQLite database couldn't be opened (e.g., `CannotOpen`, directory missing, permissions). Since all callers use fire-and-forget (`_ = _chatDb.AddMessageAsync(...)`), the exceptions were never observed, causing repeated `UnobservedTaskException` entries in `crash.log`.

Additionally, `GetConnectionAsync()` cached the `_db` field before `CreateTableAsync` succeeded — if initialization failed, the broken connection was cached permanently with no recovery path.

## Fix

- **Try/catch on all public methods**: SQLite and IO exceptions are caught and logged via `Debug.WriteLine` instead of propagating as unobserved exceptions. Methods return safe defaults on failure (-1, false, 0, empty lists).
- **Connection caching fix**: `_db` is now assigned only after successful table creation. Previously it was set before `CreateTableAsync`, so a failure would cache a broken connection.
- **Auto-recovery**: On error, the cached `_db` connection is reset to `null`, so the next call retries from scratch instead of reusing a broken connection.

## Tests

Added 11 tests in `ChatDatabaseResilienceTests.cs`:
- Valid path operations succeed and return expected values
- Invalid path operations return safe defaults without throwing
- Connection recovery works after switching from invalid to valid path
- Round-trip persistence (insert + read back)
- All error methods covered: AddMessage, BulkInsert, UpdateToolComplete, UpdateReasoning, GetAllMessages, HasMessages, GetMessageCount, ClearSession

### DIFF
--- a/PolyPilot/Services/ChatDatabase.cs
+++ b/PolyPilot/Services/ChatDatabase.cs
@@ -362,12 +362,8 @@ public class ChatDatabase : IChatDatabase
             if (Interlocked.CompareExchange(ref _db!, null!, failedConn) == failedConn)
                 try { _ = failedConn.CloseAsync(); } catch { }
         }
-        else
-        {
-            var old = _db;
-            _db = null;
-            try { _ = old?.CloseAsync(); } catch { }
-        }
+        // When failedConn is null, GetConnectionAsync already cleaned up the failed
+        // connection before throwing — _db was never set, so there is nothing to evict.
         System.Diagnostics.Debug.WriteLine($"[ChatDatabase] {method} failed: {ex.Message}");
     }
 }


StephaneDelcroix:
## 🔄 Multi-Model Re-Review (Round 3): PR #241

5-model re-review after pinning SQLite packages and fixing `ResetConnection`.

---

### Previous Findings Status

| Issue | Status |
|-------|--------|
| A — BulkInsertAsync lacks transaction | ✅ FIXED (round 1) |
| B — GetConnectionAsync race condition | ✅ FIXED (round 1) |
| C — SetDbPathForTesting doesn't reset `_db` | ✅ FIXED (round 1) |
| D — `Version="*"` in production csproj | ✅ **FIXED** — pinned to `1.9.172` / `2.1.11` |
| E — LogError leaks connection handle | ✅ FIXED (round 2) |
| F — ResetConnection doesn't close old connection | ✅ **FIXED** — same close pattern as LogError |

---

### New Consensus Findings (2+ models)

**🟡 MODERATE — `ChatDatabase.cs:138-146` — Connection leak on init failure** (3/5 models)

If `CreateTableAsync` or `ExecuteAsync("CREATE INDEX...")` throws inside `GetConnectionAsync`, the newly created `conn` is abandoned without being closed. The caller's `catch` calls `LogError`, but `LogError` closes `_db` (which is still `null`), not `conn`. On a corrupt database, every operation call leaks a file handle.

Fix: wrap the init block so `conn` is closed on failure:
```csharp
var conn = new SQLiteAsyncConnection(DbPath);
try {
    await conn.CreateTableAsync<ChatMessageEntity>();
    await conn.ExecuteAsync("CREATE INDEX IF NOT EXISTS ...");
}
catch {
    try { await conn.CloseAsync(); } catch { }
    throw;
}
_db = conn;
return conn;
```

**🟡 MODERATE — `ChatDatabase.cs:341-343` — `LogError` bypasses `_initLock`; can evict healthy connection** (3/5 models)

`LogError` reads and nulls `_db` without holding `_initLock`. Race scenario:
1. Thread A operation fails; suspended before `LogError` runs
2. Thread B calls `GetConnectionAsync`, creates fresh `conn2`, sets `_db = conn2` under lock
3. Thread A's delayed `LogError` captures `conn2` (healthy), nulls `_db`, closes it
4. Thread B and all in-flight callers now operate on a closed connection

Fix: use `Interlocked.CompareExchange` to only evict the specific connection that failed:
```csharp
private void LogError(string method, Exception ex, SQLiteAsyncConnection? failedConn = null)
{
    Interlocked.CompareExchange(ref _db, null, failedConn);
    try { _ = failedConn?.CloseAsync(); } catch { }
    Debug.WriteLine($"[ChatDatabase] {method} failed: {ex.Message}");
}
// Callers: capture db before try, pass it to LogError
```

---

### CI Status
⚠️ No CI configured. Manual: **1639/1640 pass** (1 pre-existing locale failure, unrelated).

---

### Recommendation: ⚠️ Request Changes (non-blocking)

Both findings require concurrent load to trigger. Address them (especially the `LogError` race) and this is ready to merge.

StephaneDelcroix:
## ✅ Multi-Model Re-Review (Round 4): PR #241 — Ready to Merge

5-model re-review after round-4 fixes (init connection leak + Interlocked.CompareExchange in LogError).

---

### Previous Findings Status

| Issue | Status |
|-------|--------|
| A — BulkInsertAsync lacks transaction | ✅ FIXED (round 1) |
| B — GetConnectionAsync race condition | ✅ FIXED (round 1) |
| C — SetDbPathForTesting doesn't reset `_db` | ✅ FIXED (round 1) |
| D — `Version="*"` in production csproj | ✅ FIXED (round 3) |
| E — LogError leaks connection handle | ✅ FIXED (round 2) |
| F — ResetConnection doesn't close old connection | ✅ FIXED (round 3) |
| G — Connection leak when `GetConnectionAsync` init throws | ✅ **FIXED** (round 4) |
| H — LogError can close healthy connection from concurrent thread | ✅ **FIXED** (round 4 + fixup) |

---

### Round-4 Fixes Applied

**Fix G** — `GetConnectionAsync` now wraps init in inner `try/catch`: if `CreateTableAsync` or `CREATE INDEX` throws, `conn` is closed before rethrowing. No file handle leak.

**Fix H (initial)** — `LogError` accepts `failedConn` param and uses `Interlocked.CompareExchange` to only evict `_db` if it still points to the failed connection. All 9 public methods capture `db` before the try block and pass it to `LogError`.

**Fix H (fixup, commit `a626249`)** — During 5-model re-review, 4/5 models flagged that the `else` branch (when `failedConn == null`) still had the race: it unconditionally nulled `_db` which could be a healthy connection initialized by a concurrent thread. The else branch was removed — when `GetConnectionAsync` throws, it already cleaned up `conn` internally, so `_db` was never set and there's nothing to evict.

---

### Post-Fix Re-Review Results

- G: ✅ FIXED — all 5 models confirmed
- H: ✅ FIXED — 4/4 completed models confirmed (else branch removed eliminates the residual race)

No new consensus issues found.

---

### CI Status
⚠️ No CI configured. Manual: **1644/1645 pass** (1 pre-existing locale failure `0.4` vs `0,4`, unrelated).

---

### Recommendation: ✅ Approve

All 8 issues resolved across 5 fix commits. The resilience layer is now correct:
- Transactions protect bulk inserts
- SemaphoreSlim prevents concurrent init races
- Failed connections are closed and not cached
- `LogError` uses CAS to safely evict only the specific failed connection
- Packages are pinned to tested versions
- 11 tests cover all error paths

StephaneDelcroix:
test failure addressed in #246

PR #308, opened 2026-03-08
feat: GitHub Codespace integration with SSH tunnel support

btessiau:
## GitHub Codespace Integration (Alpha)

Closes #307

### What
Connect PolyPilot to running GitHub Codespaces via SSH tunnel. Sessions in a codespace get their own sidebar group with health monitoring, auto-reconnect, and lifecycle management.

### How it works
1. **Discovery**: `gh cs list --json` finds running codespaces
2. **Connection**: SSH tunnel to the codespace's copilot headless server (port 4321)
3. **Health check**: Background loop monitors tunnel health, reconnects on failure
4. **Opt-in**: Feature is behind Settings → Alpha → Codespaces toggle (off by default)

### Architecture
- `CodespaceService` (3 files) — SSH tunnel lifecycle, codespace discovery, diagnostics
- `CopilotService.Codespace.cs` — Group management, health check loop, session resume
- Settings toggle — `CodespacesEnabled` property with runtime sync

### Requirements
- **SSH server in codespace** — Required. Without SSH there is no way to start copilot remotely or establish a tunnel. The port-forward path (`gh cs ports forward`) exists as graceful degradation but cannot start copilot — it will transition to SetupRequired state.
- `gh` CLI authenticated with codespace scope

### Key design decisions
- **CodespaceService injected via DI** (singleton), not instantiated per-call
- **Health check state writes marshaled to UI thread** via `InvokeOnUI` to prevent data races with Blazor render
- **Per-group reconnect lock** (`SemaphoreSlim`) prevents concurrent reconnect attempts from orphaning SSH tunnel processes
- **`StopCodespaceHealthCheckAsync`** uses async cancellation (no blocking `.Wait()`)
- **`sshFailed` race** fixed with `Volatile.Read/Write` on int (C# bool has no Volatile overload)
- **Null guards** on `ChangeModelAsync` and `AbortSessionAsync` for placeholder codespace sessions (Session=null until tunnel connects)
- **Toggle isolation**: when CodespacesEnabled is OFF, no placeholder sessions are created on restore, no health check starts, codespace UI is hidden
- **SDK bumped to 0.1.31** for `CopilotSession.ServerUri` property needed by codespace client routing

### Test coverage
- 113 new tests across 8 test files (CodespaceModel, Service, HealthCheck, Isolation, ClientState, ClientRouting, ClientInvariant, UX)
- All 2,218 tests pass
- Tests cover: model contracts, error messages, toggle isolation, session isolation from codespace failures
[...]

PR #315, opened 2026-03-08
fix: codespace "New Session" button silently fails

### LINKED ISSUE #314: Bug: Clicking 'New Session' on connected codespace group silently fails
## Bug Description

Clicking the "➕ New Session" button on a connected codespace group (green dot) does nothing — no new session is created, no error is shown to the user.

## Root Cause

Three independent bugs combine to create this silent failure:

### 1. Fire-and-forget async lambda (silent exception swallowing)

The `@onclick` handler in `SessionSidebar.razor` uses a **void-returning lambda** that calls an `async Task` method:

```csharp
@onclick="() => { openGroupMenuId = null; QuickCreateSessionForCodespace(gId); }"
```

In Blazor, `@onclick` with a void delegate (`Action`) does NOT await the returned `Task`. When `QuickCreateSessionForCodespace` hits its first `await` (calling `CreateSessionAsync`), it yields — and Blazor has no reference to the Task. Any exception thrown during the async portion is silently swallowed as an unobserved Task exception. The UI never re-renders after the async work completes or fails.

### 2. Error message displayed in wrong UI section

`QuickCreateSessionForCodespace` correctly catches exceptions and stores them in `createError`, but this variable is only rendered in the `CreateSessionForm` component at the **top of the sidebar** — not in the codespace section where the user actually clicked. Even with Bug 1 fixed, the error would appear in a completely different part of the UI.

### 3. Health check does not detect dead remote copilot process

The codespace health check verifies:
- SSH tunnel process is alive (`tunnel.IsAlive`) ✅
- CopilotClient exists in dictionary (`_codespaceClients.ContainsKey`) ✅

But it never checks whether the **remote copilot process is still listening** on the tunnel port. An SSH tunnel can stay alive long after the remote `copilot --headless` process dies (e.g., codespace went to sleep and came back, copilot was killed manually). The green "connected" dot persists, the "New Session" button appears, but every `CreateSessionAsync` call fails because the HTTP request through the tunnel gets connection-refused.

## Expected Behavior

1. Clicking "New Session" on a connected codespace group should either create the session or show an error **in the codespace section**
2. The health check should detect when the remote copilot dies behind a live tunnel and transition to reconnecting state
3. Error messages from codespace operations should be visible to the user

## Fix

1. **Async lambda**: Change to `async () => { ... await QuickCreateSessionForCodespace(gId); }` so Blazor properly awaits the Task and re-renders
2. **Error display**: Show `createError` in the codespace group section (below sessions)
3. **TCP probe**: Add a TCP connection probe in the health check — when tunnel is alive and client exists, verify the remote port is actually reachable before marking the group as healthy

## Affected Files

- `PolyPilot/Components/Layout/SessionSidebar.razor` — void lambda, error display
- `PolyPilot/Components/Layout/SessionSidebar.razor.css` — error styling
- `PolyPilot/Services/CopilotService.Codespace.cs` — health check TCP probe

## Related

- PR #308 (codespace integration feature)

btessiau:
## Summary

Fixes #314 — Clicking "➕ New Session" on a connected codespace group (green dot) silently does nothing.

> **Note:** This PR depends on #308 (codespace integration). Merge #308 first, then this cleanly applies on top.

## Root Cause

Three independent bugs combine to create this silent failure:

| # | Bug | Impact |
|---|-----|--------|
| 1 | **Fire-and-forget async lambda** — `@onclick` handler is void-returning, so Blazor never awaits `QuickCreateSessionForCodespace` | Exception silently swallowed |
| 2 | **Error in wrong UI section** — `createError` only renders in `CreateSessionForm` at top of sidebar, not codespace section | Error invisible to user |
| 3 | **Stale client undetected** — Health check only checks `tunnel.IsAlive` + `ContainsKey()`, never probes remote port | Green dot stays after remote copilot dies |

### `SessionSidebar.razor`
- Changed void lambda to `async () => { ... await QuickCreateSessionForCodespace(...) }` so Blazor properly awaits and re-renders
- Added `codespace-create-error` banner below codespace group sessions (click to dismiss)

### `SessionSidebar.razor.css`
- Added `.codespace-create-error` style

### `CopilotService.Codespace.cs`
- Health check now does a TCP probe to the tunnel's local port when tunnel + client + Connected all pass
- If probe fails: disposes stale client, removes from `_codespaceClients`, falls through to reconnect path
- The green dot will correctly transition to reconnecting when the remote copilot dies

### `CodespaceSessionCreationTests.cs` (new, 8 tests)
- Missing client → descriptive error (not "Service not initialized")
- No session state leaked on failure
- Non-codespace groups skip codespace guard
- TCP probe detects stale client (old code would skip)
- Healthy group still skips correctly
- All disconnected states block creation
- Connected state allows creation
- Error message mentions health check + retry guidance

## Test Results

2,223 pass (8 new), 3 pre-existing CLI binary failures (unrelated).

## Risk

- **TCP probe overhead**: One extra TCP connect per health check cycle (15s) per connected codespace group. Sub-millisecond for localhost.
- **No behavior change for healthy connections**: Probe succeeds instantly → `continue` as before.
- **Backward compatible**: No API changes.

### DIFF
--- a/PolyPilot/Services/CopilotService.Codespace.cs
+++ b/PolyPilot/Services/CopilotService.Codespace.cs
@@ -389,9 +389,13 @@ public partial class CopilotService : IAsyncDisposable
                 try
                 {
                     using var probe = new System.Net.Sockets.TcpClient();
-                    await probe.ConnectAsync("127.0.0.1", tunnel.LocalPort, ct);
+                    using var probeCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
+                    probeCts.CancelAfter(TimeSpan.FromSeconds(3));
+                    await probe.ConnectAsync("127.0.0.1", tunnel.LocalPort, probeCts.Token);
                     portReachable = true;
                 }
+                catch (OperationCanceledException) when (!ct.IsCancellationRequested)
+                { /* per-probe timeout — treat as unreachable */ }
                 catch (OperationCanceledException) { throw; }
                 catch { /* TCP connect failed — remote end not listening */ }
 


btessiau:
Thanks for the review! Addressed both findings in `1109976`:

**🟡 TCP probe comment** — Added a 3-line comment explaining why the bare `TcpClient` probe suffices for SSH tunnels (they refuse connections outright when the remote end dies), while `CodespaceService.IsCopilotListeningAsync` is needed for non-SSH tunnels like `gh cs ports forward` which accept locally regardless.

**🟢 Shared `createError`** — Split into separate `codespaceCreateError` variable so codespace errors only show in the codespace section and regular session errors only show in `CreateSessionForm`. No cross-section flashing.

Also resolved merge conflicts with `origin/main`. 2,226 tests pass.

btessiau (commented):
## Re-review after `1109976`

Both suggestions from my first review are addressed:

✅ **TCP probe comment** — Clear explanation of why bare `TcpClient.ConnectAsync` suffices for SSH tunnels vs. the `IsCopilotListeningAsync` read-probe needed for `gh cs ports forward`
✅ **Error variable isolation** — `codespaceCreateError` is now separate from `createError`, preventing cross-section error flashing

Merge conflict with #311 (collapsible unpinned sessions) resolved cleanly — codespace UI section repositioned correctly.

One housekeeping note: `review.md` was accidentally committed in #308 and is now on `main`. Not part of this PR's diff, but should be removed in a follow-up.

No new issues. LGTM 👍

PR #318, opened 2026-03-09
feat: Intercept SDK task tool for multi-agent orchestration (#229)

### LINKED ISSUE #229: Rework multi-agent orchestrator: intercept SDK task tool to surface sub-agents as PolyPilot sessions
## Problem

The OrchestratorReflect mode relies on the orchestrator producing `@worker:name` text blocks, which PolyPilot's reflect loop parses via regex and dispatches to worker sessions. However, the orchestrator is a full Copilot SDK session with built-in tools (`task`, `grep`, `edit`, etc.), and the model **strongly prefers using the SDK's `task` sub-agent tool** over producing text-based `@worker:` blocks.

This causes:
1. The orchestrator does all work itself via SDK sub-agents instead of delegating to PolyPilot worker sessions
2. Worker sessions sit idle with zero events
3. The reflect loop's `ParseTaskAssignments` finds no `@worker:` blocks and stalls
4. Sub-agent work is invisible in the PolyPilot UI — no progress tracking, no session history

### Current workaround (PR #225)
Rewrote the orchestrator's planning prompt to forcefully say "You are a DISPATCHER ONLY, you do NOT have tools" and removed tool-like language from the RoutingContext. This works but fights against the model's natural behavior.

## Proposed Approach: Intercept SDK `task` Tool

Instead of fighting the model, **embrace its preference for the `task` tool** and intercept it to route through PolyPilot's worker sessions.

### Architecture

1. **Disable the SDK's built-in `task` tool** via `ExcludedTools = new List<string> { "task" }` on the orchestrator's `SessionConfig`

2. **Register a custom `AIFunction`** (e.g., `delegate_to_worker`) that:
   - Receives the prompt and agent_type from the model
   - Maps to the appropriate PolyPilot worker session (or creates one dynamically)
   - Calls `ExecuteWorkerAsync(workerName, task, originalPrompt, ct)` — the same code path the reflect loop already uses
   - Returns the worker's response back to the SDK as the tool result

3. **Workers become visible sessions** — each sub-agent delegation shows up as a real PolyPilot session with:
   - Full event stream (tool calls, edits, file changes)
   - Progress indicators in the UI
   - Worktree isolation support
   - Persistent history

### Key Research Findings

- **`SessionConfig.Tools`** accepts `List<AIFunction>` — custom tools added alongside built-in ones (see `ShowImageTool` pattern)
- **`SessionConfig.ExcludedTools`** and **`SessionConfig.AvailableTools`** exist in SDK 0.1.26 but aren't used by PolyPilot yet
- **`subagent.started`** events contain `toolCallId`, `agentName`, `agentDisplayName`
- **`tool.execution_start`** for `task` contains the full `prompt`, `agent_type`, and `description` — all the data needed to route to a worker
- The model naturally names its sub-agent calls descriptively (e.g., "Implement steering messages", "Challenge steering implementation") — these could become session display names

### Open Questions

- Does `ExcludedTools` work for the built-in `task` tool?
- Can a custom `AIFunction` be named `task` to transparently replace the built-in, or does it need a different name?
- Should dynamic worker sessions be created on-the-fly, or pre-allocated from the preset?
- How to handle the `explore` agent type — map to a lightweight read-only worker?
- What happens to `general-purpose` sub-agents that the orchestrator spawns for complex multi-step work?

### Benefits

- **Stop fighting the model** — let it use tools naturally
- **Full visibility** — every sub-agent becomes a trackable session
- **Worktree isolation** — workers can have their own git worktrees
- **Persistence** — sub-agent work survives app restarts
- **Simplify the reflect loop** — no more `@worker:` text parsing, `ParseTaskAssignments` regex, or prompt engineering to force text dispatch

### Files Involved

- `PolyPilot/Services/CopilotService.cs` — `SessionConfig.Tools`, `SessionConfig.ExcludedTools`
- `PolyPilot/Services/CopilotService.Organization.cs` — `ExecuteWorkerAsync`, `SendViaOrchestratorReflectAsync`
- `PolyPilot/Services/ShowImageTool.cs` — existing custom `AIFunction` pattern to follow
- `PolyPilot/Models/ModelCapabilities.cs` — preset definitions, RoutingContext

PureWeen:
Closes #229

## What

Replaces fragile @worker: text-parsing with structured tool-call dispatch for multi-agent orchestration.

### Changes
- Block SDK built-in task tool via ExcludedTools=["task"]
- Register custom task AIFunction that routes to workers via ExecuteWorkerAsync
- Thread-safe result collection (WorkerDelegationContext)
- Tool-dispatch path with @worker: text fallback
- Reconnect handling (clear _reflectToolConfigured)
- Codespace client support (GetClientForGroup)
- 14 unit tests

### Testing
- 2114+ tests passing
- Use Implement & Challenge preset (OrchestratorReflect mode) to verify tool-dispatch
- Fallback to @worker: text parsing works when tool setup fails

### DIFF
--- a/PolyPilot.Tests/ChatExperienceSafetyTests.cs
+++ b/PolyPilot.Tests/ChatExperienceSafetyTests.cs
@@ -846,8 +846,10 @@ public class ChatExperienceSafetyTests
         Assert.True(sessionNotFoundIdx > 0);
 
         var afterNotFound = source.Substring(sessionNotFoundIdx, Math.Min(2000, source.Length - sessionNotFoundIdx));
-        Assert.Contains("McpServers", afterNotFound);
-        Assert.Contains("SkillDirectories", afterNotFound);
+        // After extracting BuildFreshSessionConfig helper, the reconnect path calls the helper
+        // instead of inlining McpServers/SkillDirectories. Verify the helper is called here
+        // and that it contains MCP/Skills (checked via helper body elsewhere).
+        Assert.Contains("BuildFreshSessionConfig", afterNotFound);
     }
 
     // =========================================================================
--- a/PolyPilot.Tests/ConnectionRecoveryTests.cs
+++ b/PolyPilot.Tests/ConnectionRecoveryTests.cs
@@ -274,67 +274,72 @@ public class ConnectionRecoveryTests
     [Fact]
     public void SendPromptAsync_FreshSessionConfig_IncludesMcpServers()
     {
-        // STRUCTURAL REGRESSION GUARD: The "Session not found" fallback must assign
-        // McpServers in the freshConfig so MCP tools survive reconnection.
+        // STRUCTURAL REGRESSION GUARD: The BuildFreshSessionConfig helper must assign
+        // McpServers so MCP tools survive reconnection and fresh session creation.
         var source = File.ReadAllText(Path.Combine(GetRepoRoot(), "PolyPilot", "Services", "CopilotService.cs"));
 
-        // Anchor on the freshConfig initializer inside the "Session not found" reconnect path
-        var freshConfigIndex = source.IndexOf("freshConfig = new SessionConfig");
-        Assert.True(freshConfigIndex > 0, "Could not find freshConfig in reconnect path");
+        // Verify BuildFreshSessionConfig helper exists and contains McpServers
+        var helperIndex = source.IndexOf("private SessionConfig BuildFreshSessionConfig(");
+        Assert.True(helperIndex > 0, "Could not find BuildFreshSessionConfig helper method");
 
-        // Extract the config block (generously sized to cover all fields)
-        var endIndex = Math.Min(freshConfigIndex + 600, source.Length);
-        var configBlock = source.Substring(freshConfigIndex, endIndex - freshConfigIndex);
-        Assert.Contains("McpServers = ", configBlock);
+        var endIndex = Math.Min(helperIndex + 2200, source.Length);
+        var helperBlock = source.Substring(helperIndex, endIndex - helperIndex);
+        Assert.Contains("McpServers = ", helperBlock);
+
+        // Verify the reconnect "Session not found" path calls the helper
+        var reconnectIndex = source.IndexOf("Session expired server-side");
+        Assert.True(reconnectIndex > 0, "Could not find reconnect expired-session path");
+        var reconnectBlock = source.Substring(reconnectIndex, Math.Min(500, source.Length - reconnectIndex));
+        Assert.Contains("BuildFreshSessionConfig", reconnectBlock);
     }
 
     [Fact]
     public void SendPromptAsync_FreshSessionConfig_IncludesSkillDirectories()
     {
-        // STRUCTURAL REGRESSION GUARD: The "Session not found" fallback must assign
-        // SkillDirectories in the freshConfig so skills survive reconnection.
+        // STRUCTURAL REGRESSION GUARD: The BuildFreshSessionConfig helper must assign
+        // SkillDirectories so skills survive reconnection and fresh session creation.
         var source = File.ReadAllText(Path.Combine(GetRepoRoot(), "PolyPilot", "Services", "CopilotService.cs"));
 
-        var freshConfigIndex = source.IndexOf("freshConfig = new SessionConfig");
-        Assert.True(freshConfigIndex > 0, "Could not find freshConfig in reconnect path");
+        var helperIndex = source.IndexOf("private SessionConfig BuildFreshSessionConfig(");
+        Assert.True(helperIndex > 0, "Could not find BuildFreshSessionConfig helper method");
 
-        var endIndex = Math.Min(freshConfigIndex + 600, source.Length);
-        var configBlock = source.Substring(freshConfigIndex, endIndex - freshConfigIndex);
-        Assert.Contains("SkillDirectories = ", configBlock);
+        var endIndex = Math.Min(helperIndex + 2200, source.Length);
+        var helperBlock = source.Substring(helperIndex, endIndex - helperIndex);
+        Assert.Contains("SkillDirectories = ", helperBlock);
     }
 
     [Fact]
     public void SendPromptAsync_FreshSessionConfig_IncludesSystemMessage()
     {
-        // STRUCTURAL REGRESSION GUARD: The "Session not found" fallback must include
+        // STRUCTURAL REGRESSION GUARD: The BuildFreshSessionConfig helper must include
         // SystemMessage so the session retains its system prompt after reconnection.
         var source = File.ReadAllText(Path.Combine(GetRepoRoot(), "PolyPilot", "Services", "CopilotService.cs"));
 
-        var freshConfigIndex = source.IndexOf("freshConfig = new SessionConfig");
-        Assert.True(freshConfigIndex > 0, "Could not find freshConfig in reconnect path");
+        var helperIndex = source.IndexOf("private SessionConfig BuildFreshSessionConfig(");
+        Assert.True(helperIndex > 0, "Could not find BuildFreshSessionConfig helper method");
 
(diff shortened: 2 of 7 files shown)


PureWeen:
## PR #318 — 5-Model Consensus Review

**Title**: feat: Intercept SDK task tool for multi-agent orchestration  
**CI**: ⚠️ No checks configured on `feature/intercept-task-tool-229`  
**Prior reviews**: None  
**Models**: claude-opus-4.6 ×2, gemini-3-pro-preview, gpt-5.3-codex (4/5 complete; 1 timed out — findings below reflect 4-model consensus)

---

### 🔴 CRITICAL — Reconnect-skip path leaves `SendingFlag` set → session deadlock (4/4 consensus)

**`PolyPilot/Services/CopilotService.cs` — reconnect-skip block**

The new orchestrator reconnect-skip path sets `IsProcessing = false` but does **not** call `Interlocked.Exchange(ref state.SendingFlag, 0)`.

`SendPromptAsync` sets `SendingFlag = 1` at entry. All other `IsProcessing = false` paths release it (either via `CompleteResponse` or explicit reset). This path does neither. The next call to `SendPromptAsync` for this session will fail the `CompareExchange(ref state.SendingFlag, 1, 0)` check and throw "Session is already processing a request" — permanently deadlocking the session until the app restarts.

**Additional missing INV-1 cleanup in the same block** (4/4 consensus):
- `FlushCurrentResponse(state)` not called before `IsProcessing = false` → any streaming content accumulated before reconnect is silently dropped
- `ClearPermissionDenials()` not called → stale permission denials persist into next turn
- `OnSessionComplete` not fired → external subscribers never unblocked

```csharp
// Missing from the reconnect-skip block:
FlushCurrentResponse(state);                             // before IsProcessing = false
Interlocked.Exchange(ref state.SendingFlag, 0);          // ← CRITICAL: prevents deadlock
state.Info.ClearPermissionDenials();
OnSessionComplete?.Invoke(sessionName);
// Also missing diagnostic log tag: [RECONNECT-SKIP] or similar
```

---

### 🟡 MODERATE — `OnStateChanged?.Invoke()` in reconnect-skip path fires on background thread (4/4 consensus)

**`PolyPilot/Services/CopilotService.cs` — same reconnect-skip block**

`OnStateChanged?.Invoke()` is called directly without `InvokeOnUI()`. This code runs in `SendPromptAsync`'s catch block, which for worker sessions executes on a `Task.Run` background thread. Project invariant INV-2 requires ALL `IsProcessing` state mutations and `OnStateChanged` notifications to be marshaled to the UI thread. Every other termination path uses `InvokeOnUI()` or `_syncContext.Post()`. This will cause cross-thread Blazor render exceptions under load.

**Fix**: Wrap the entire block in `InvokeOnUI(() => { ... })`.

---

### 🟡 MODERATE — Escalating poll calls `CompleteResponse` on background thread when `_syncContext == null` (4/4 consensus)

**`PolyPilot/Services/CopilotService.Organization.cs` — `EnsureOrchestratorToolsAsync`, `OnToolDispatchEnd` lambda**

```csharp
if (_syncContext != null)
    _syncContext.Post(_ => CompleteResponse(cur, endGen), null);
else
    CompleteResponse(cur, endGen);  // ← called on Task.Run thread
```

The `else` branch calls `CompleteResponse` directly on the thread pool. `CompleteResponse` mutates `IsProcessing`, fires `OnStateChanged`, and clears `SendingFlag` — all must run on the UI thread per INV-2. The watchdog uses `InvokeOnUI()` (which handles null `_syncContext` safely); this code should too. In production MAUI `_syncContext` is typically non-null, but the fallback path is a correctness bug.

**Fix**: Replace `else CompleteResponse(cur, endGen)` with `else InvokeOnUI(() => CompleteResponse(cur, endGen))`.

---

### 🟡 MODERATE — `AddPendingTask` silently overwrites prior task for same worker name (3/4 consensus)

**`PolyPilot/Services/WorkerDelegationTool.cs` — `AddPendingTask`**

```csharp
internal void AddPendingTask(string workerName, Task<ToolDispatchedResult> task)
{
    _pendingTasks[workerName] = task;  // overwrites silently
}
```

The round-robin index wraps around when all workers have been dispatched. With 2 workers and 3 dispatches, the sequence is w1→w2→w1. The second w1 dispatch overwrites the first task in `_pendingTasks`. The first w1 worker continues running in the background but its result is permanently lost — `AwaitAllPendingAsync` only collects the second task. This also produces an unobserved task exception if the first task faults.

**Fix**: Use a list-of-tasks keyed by worker name, or append a dispatch counter to the key.

---

### 🟡 MODERATE — Stale pending tasks from interrupted dispatch survive into new context (3/4 consensus)

**`PolyPilot/Services/CopilotService.Organization.cs` — `EnsureOrchestratorToolsAsync`**

```csharp
var context = _delegationContexts.GetOrAdd(orchestratorName, _ => new WorkerDelegationContext());
```

`GetOrAdd` reuses the existing `WorkerDelegationContext` on reconnect. The orchestration loops call `FullReset` at the start of a new cycle (which clears `_pendingTasks`), but the reflect loop only calls `Reset`/`ResetResults` — which deliberately preserve pending tasks. After a reconnect mid-reflect-loop, the new iteration's `EnsureOrchestratorToolsAsync` reuses the context with surviving pending tasks from the interrupted dispatch. Those tasks are then awaited by `AwaitAllPendingAsync` in the new iteration, mixing stale results from the old dispatch with new ones.

---

### 🟡 MODERATE — NRE in synthesis session creation when `_client` is null (2/4 consensus)

**`PolyPilot/Services/CopilotService.Organization.cs` — synthesis path**

```csharp
GitHub.Copilot.SDK.CopilotClient synthClient;
try { synthClient = GetClientForGroup(orchestratorGroupId2); }
catch { synthClient = _client; }
var freshSession = await synthClient.CreateSessionAsync(...);  // NRE if _client is null
```

If `GetClientForGroup` throws (codespace tunnel disconnected) AND `_client` is null (after a failed reconnect set `_client = null; IsInitialized = false`), `synthClient` is null and the next line throws `NullReferenceException`, crashing the synthesis phase with an opaque error.

**Fix**: `catch { synthClient = _client ?? throw new InvalidOperationException("No client available for synthesis session"); }`

---

### Test Coverage Assessment

The 14 new tests in `WorkerDelegationToolTests.cs` are comprehensive for `WorkerDelegationContext` behavior (concurrency, timeouts, round-robin, nudge scenarios). The 6 new tests in `MultiAgentRegressionTests.cs` use the source-text-as-spec pattern (read source file, assert string patterns exist) which is fragile but consistent with the existing regression test style.

**Missing test coverage for new code paths**:
- Reconnect-skip path: no test verifying `SendingFlag` is cleared, `FlushCurrentResponse` called, and `OnStateChanged` fires on UI thread
- `AddPendingTask` overwrite: the existing `Context_ConcurrentAddResult_DoesNotCorrupt` test doesn't cover the overwrite scenario (same worker dispatched twice → second overwrites first)
- Stale pending task contamination after reconnect: no test for `GetOrAdd` reuse with surviving tasks

**Suggested test cases**:
1. `ReconnectSkipPath_ClearsSendingFlag` — verify `SendingFlag == 0` after reconnect-skip fires
2. `AddPendingTask_SameWorkerTwice_BothTasksAwaited` — currently only the second task is collected
3. `AwaitAllPendingAsync_StaleTasksFromPreviousCycle_AreNotIncluded`

---

### Summary

The thundering-herd fix (`_clientReconnectLock`) and `GetClientForGroup` usage in `EnsureOrchestratorToolsAsync` are correct. The `WorkerDelegationContext` thread-safety model is sound. The core issue is that the new reconnect-skip path introduces a new `IsProcessing = false` path that violates 4 of the project's 12 required cleanup steps — most critically, `SendingFlag` is not released, which will deadlock sessions.

⚠️ **Request changes** — address the CRITICAL SendingFlag issue + the two threading violations before merge.

---
*4-model review: claude-opus-4.6 ×2, gemini-3-pro-preview, gpt-5.3-codex*

PR #324, opened 2026-03-09
fix: relaunch.sh — prefer ~/.dotnet SDK and skip Xcode version check

btessiau:
## Problem

`relaunch.sh` fails on machines where:

1. **.NET 10 SDK is installed via `dotnet-install.sh`** — installs to `~/.dotnet/`, but the system `dotnet` at `/usr/local/share/dotnet/` is an older version (e.g. 7.0). The script's bare `dotnet build` picks the wrong one.

2. **Xcode is a minor version ahead** of what the .NET MacCatalyst SDK expects (e.g. Xcode 26.3 vs expected 26.2), causing the build to fail with a version mismatch error.

## Fix

- **PATH**: Prepend `~/.dotnet` to PATH if `~/.dotnet/dotnet` exists. Conditional — no-op on machines with a single system-wide install.
- **Xcode**: Pass `-p:ValidateXcodeVersion=false` to `dotnet build`. Apple ships Xcode updates faster than .NET SDK updates, so this mismatch is common.

### DIFF
--- a/PolyPilot/relaunch.sh
+++ b/PolyPilot/relaunch.sh
@@ -30,7 +30,9 @@ OLD_PIDS=$(ps -eo pid,comm | grep "PolyPilot" | grep -v grep | grep -v "PolyPilo
 echo "🔨 Building..."
 cd "$PROJECT_DIR"
 
-# -p:ValidateXcodeVersion=false handles Xcode minor version mismatches
+# -p:ValidateXcodeVersion=false bypasses the .NET SDK's Xcode version-string gate.
+# Safe for minor version skew (Apple ships Xcode faster than .NET certifies it).
+# A major Xcode incompatibility will still surface as a compile/link error.
 BUILD_OUTPUT=$(dotnet build PolyPilot.csproj -f net10.0-maccatalyst -p:ValidateXcodeVersion=false 2>&1)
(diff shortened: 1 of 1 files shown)


btessiau (commented):
## PR #324 Review -- Multi-Model Consensus Report

**CI Status:** ⚠️ No checks configured
**Models used:** claude-opus-4.6 ×2, claude-sonnet-4.6, gemini-3-pro-preview, gpt-5.3-codex

---

### ✅ PATH prepend is correct
The `[ -x "$HOME/.dotnet/dotnet" ]` guard, timing (before first `dotnet` invocation), and process-local `export` are all sound. No issues found.

### 🟡 MODERATE -- `ValidateXcodeVersion=false` was unconditional with no comment (3/5 models flagged)
**`PolyPilot/relaunch.sh:33`** -- The flag correctly silences minor Xcode version skew (the stated intent), but also silences major version incompatibilities. Since `relaunch.sh` kills the old app before launching the new build, a silently broken binary is more disruptive here than a clear build error.

**Status: ✅ FIXED** -- A 3-line comment block has been added documenting the tradeoff. All 5 models confirmed the fix in re-review.

---

### Recommendation: ✅ Approve
All previous findings have been addressed. The code changes are correct and well-scoped.

PR #358, opened 2026-03-11
Fix relaunch.ps1 build output path

PureWeen:
The relaunch.ps1 script was failing with 'file not found' when attempting to launch the built executable. The issue was in the $BuildDir path construction on line 13, which was missing the win-x64 runtime identifier (RID) subdirectory.

**The Problem:**
When building a Windows Forms .NET 10 application with RID-specific output, the build artifacts are placed in a subdirectory matching the RID (e.g., win-x64). The original path was:
\\\
bin\Debug\net10.0-windows10.0.19041.0
\\\

But the actual executable location is:
\\\
bin\Debug\net10.0-windows10.0.19041.0\win-x64
\\\

**The Fix:**
Updated the path to include the win-x64 RID subdirectory so Start-Process can correctly locate and launch the built exe.

### DIFF
--- a/PolyPilot/relaunch.ps1
+++ b/PolyPilot/relaunch.ps1
@@ -1,3 +1,7 @@
+param(
+    [string]$Configuration = 'Debug'
+)
+
 # Builds PolyPilot, launches a new instance, waits for it to be ready,
 # then kills the old instance(s) for a seamless handoff.
 #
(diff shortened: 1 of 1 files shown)


PureWeen (commented):
## PR #358 Round 2 -- Approve

Finding 1 (arbitrary RID selection): FIXED -- now filters by exe presence.
Finding 2 (non-RID fallback): Still present but LOW risk for .NET 10 Desktop.

Verdict: Approve -- code is materially better.

*4-model AI consensus (Round 2)*

PR #373, opened 2026-03-13
fix: session stability hardening

PureWeen:
## Problem

Sessions die instantly after reconnect due to compounding bugs:

1. **Corrupted session files kill sessions permanently** - When events.jsonl contains tool output with words like "ephemeral", the CLI JSON parser chokes on session.resume. No catch existed, so the session dies.
2. **Orphaned event handlers race with new state** - Stale SessionIdleEvent callbacks from the old session clear IsProcessing on the shared Info object, killing the new session. Each reconnect stacks another handler.
3. **Watchdog crashes leave sessions permanently stuck** - If watchdog cleanup throws, IsProcessing was never cleared.
4. **posix_spawn failures not detected** - Shell failures did not trigger auto-recovery.
5. **relaunch.sh kills copilot server** - Grep pattern matched the bundled copilot binary too.

## Fixes

- **Corrupted session file recovery**: Catch "corrupted" in resume, create fresh session
- **IsOrphaned flag**: HandleSessionEvent and CompleteResponse skip orphaned states
- **Generation invalidation**: Old state ProcessingGeneration set to long.MaxValue
- **Watchdog crash safety**: Outer catch clears IsProcessing (INV-1 compliant)
- **Shell failure detection**: posix_spawn treated like permission denial for recovery
- **relaunch.sh**: End-of-line anchor to only match app binary

## Testing
2549 tests pass, 5 pre-existing failures (font scaling + DevTunnel - unrelated)

### DIFF
--- a/PolyPilot/Services/CopilotService.cs
+++ b/PolyPilot/Services/CopilotService.cs
@@ -2648,6 +2648,8 @@ ALWAYS run the relaunch script as the final step after making changes to this pr
                                             if (otherMeta?.GroupId != null &&
                                                 groupSnapshots.Any(g => g.Id == otherMeta.GroupId && g.IsCodespace))
                                                 continue;
+                                            // Check cancellation between siblings for clean shutdown
+                                            if (cancellationToken.IsCancellationRequested) break;
                                             try
                                             {
                                                 var settings = _currentSettings ?? ConnectionSettings.Load();
@@ -2665,7 +2667,7 @@ ALWAYS run the relaunch script as the final step after making changes to this pr
                                                 if (!string.IsNullOrEmpty(otherState.Info.WorkingDirectory))
                                                     cfg.WorkingDirectory = otherState.Info.WorkingDirectory;
                                                 var resumed = await newClient.ResumeSessionAsync(
-                                                    otherState.Info.SessionId, cfg, CancellationToken.None);
+                                                    otherState.Info.SessionId, cfg, cancellationToken);
                                                 // Re-check after await — a concurrent SendPromptAsync
                                                 // may have started processing while we were resuming.
                                                 // Orphan the just-resumed session rather than cancel a live turn.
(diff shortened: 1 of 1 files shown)


PureWeen:
## Re-Review: fix/session-stability-hardening

> Multi-model re-review (5/5 models) — updated after 5 additional commits

### Previous Findings Status

| # | Finding | Status |
|---|---------|--------|
| 1 | 🔴 TOCTOU race in sibling re-resume | ✅ FIXED (post-await re-check added) |
| 2 | 🟡 Missing MCP servers in sibling config | ✅ FIXED (`LoadMcpServers` + `LoadSkillDirectories` added) |
| 3 | 🟡 Watchdog crash loses `CurrentResponse` | ✅ FIXED (`crashResponse = FlushedResponse + CurrentResponse`) |
| 4 | 🟡 Broad `"session file"` exception filter | ✅ FIXED (tightened to `"session file is"` + `"Invalid literal value"`) |
| 5 | 🟢 Indentation corruption in `TriggerToolHealthRecovery` | ❌ STILL PRESENT |

---

## New Finding (Consensus: 3/5 models)

### 🟡 MODERATE — Leaked  in sibling re-resume abandon paths
**File:**  (sibling loop, ~lines 2672 and 2703)

Two  paths in the sibling re-resume loop abandon a live   without disposing it:

**Path 1 — post-await TOCTOU guard:**
```csharp
var resumed = await newClient.ResumeSessionAsync(otherState.Info.SessionId, cfg, CancellationToken.None);
if (otherState.Info.IsProcessing)
{
    Debug($"[RECONNECT] Sibling started processing during re-resume — skipping");
    continue;  // ← 'resumed' is never disposed
}
```

**Path 2 —  failure (rapid back-to-back reconnects):**
```csharp
{
    siblingState.IsOrphaned = true;
    continue;  // ← 'resumed' is never disposed
}
```

In both cases,  has already created a server-side session with an open transport (stdio pipe or TCP connection to the copilot CLI process). Without , these handles accumulate on rapid reconnects or in multi-session workspaces. Under flapping network conditions with multiple sessions, this will exhaust file descriptors or produce stale server-side sessions.

**Fix:**
```csharp
// Path 1
if (otherState.Info.IsProcessing)
{
    Debug($"[RECONNECT] Sibling started processing during re-resume — skipping");
    try { await resumed.DisposeAsync(); } catch { }
    continue;
}

// Path 2
{
    siblingState.IsOrphaned = true;
    try { await resumed.DisposeAsync(); } catch { }
    continue;
}
```

---

## Remaining Minor Issue

** indentation** () — the  call still has 24 extra leading spaces in the updated diff. Functionally harmless in C# but creates misleading visual structure.

---

## Recommended Action

⚠️ **Request changes** (one remaining item)

All four major findings from the initial review are fixed. One new 🟡 MODERATE issue: the two  CopilotSession leak paths need . This is a straightforward two-line fix. Once addressed the PR is ready to merge.

Issue #365, opened 2026-03-12
Watchdog Case B force-completes worker sessions prematurely when SDK pauses mid-turn


PureWeen:
## Summary

Multi-agent orchestration worker sessions can be force-completed by the watchdog while the SDK is still actively processing, causing **truncated results** sent to the orchestrator and **orphaned streaming** visible in the UI.

## Reproduction

Observed with "PR Review Squad" — worker-3 was processing normally, then the SDK stopped sending events for ~6 minutes. The watchdog:

1. **21:18:44** — Last SDK event received (`TurnEnd`/`TurnStart` pair, `IsProcessing=True`)
2. **21:21:58** (194s later) — Watchdog Case B fires. `events.jsonl` is still fresh → **deferred** (1/40)
3. **21:24:58** (180s after deferral) — Case B fires again. `events.jsonl` age now exceeds 300s → **force-completes** with only **166 chars** of `FlushedResponse`
4. **21:24:58** — Orchestrator receives the 166-char truncated result, immediately sends synthesis prompt
5. **21:25:10** — Orchestrator completes with 964-char synthesis (based on incomplete worker input)
6. **21:25:41 → 21:32:58** — SDK **resumes streaming** events to worker-3 with `IsProcessing=False`. User sees messages appearing but session is not marked busy.
7. **21:32:58** — `SessionIdleEvent` finally arrives, `CompleteResponse` skips ("IsProcessing already false")

## User-visible symptoms

- Worker session shows streaming messages but no "busy" indicator
- Orchestrator synthesis is based on truncated/incomplete worker output
- Worker content generated after force-completion is lost (appended to history but never collected)

## Root cause

The watchdog Case B freshness check (`events.jsonl` age < 300s) is designed to detect "tool finished but `SessionIdleEvent` never arrived" (SDK bug #299). However, the SDK can legitimately pause event delivery for **longer than 300s** during complex tool executions (e.g., long `bash` or `task` tool runs server-side).

When `events.jsonl` ages past 300s, Case B concludes the session is done and calls `CompleteResponse`. But the SDK then resumes sending events — the session was actually still running server-side. The TCS resolves early, the orchestrator collects truncated output, and all subsequent content is orphaned.

## Key code paths

- **Watchdog Case B**: `CopilotService.Events.cs` ~line 1926-1989
- **`caseBEventsActive` check**: line 1960 — `lastWrite > startedAt.Value && age < 300`
[...]

PR #426, opened 2026-03-24
fix: detect dead connections in watchdog Case B via file size growth check

PureWeen:
## Problem
Multi-agent worker sessions get stuck in `IsProcessing=true` for up to **30 minutes** when the JSON-RPC connection is lost (`ConnectionLostException`). The watchdog Case B deferral logic uses events.jsonl modification time to detect if the CLI is still active, but the 1800s multi-agent freshness window means a file modified before the connection died still appears "fresh" for 30 minutes.

### Observed symptoms
- Sessions show "Thinking..." or spinner stuck for 20+ minutes
- Debug log: `[WATCHDOG] Case B deferred — events.jsonl modified since turn start, session still active (elapsed=120s, totalProcessing=1455s, deferral=12/40, freshness=1800s [multi-agent])`
- crash.log shows `ConnectionLostException: The JSON-RPC connection with the remote party was lost`

## Root Cause
When a ConnectionLostException kills the JSON-RPC connection:
1. No more SDK events arrive → watchdog inactivity timer fires after 120s
2. Case B checks events.jsonl freshness: file was written before connection died, passes `age < 1800s`
3. Each deferral resets `LastEventAtTicks`, creating a 120s cycle
4. Session stays stuck until file age exceeds 1800s (~30 min) or 40-deferral cap (~80 min)

## Fix
Added a **file size growth check** to Case B deferral logic:
- Track events.jsonl file size at each Case B check (`WatchdogCaseBLastFileSize`)
- If the file hasn't grown across 2 consecutive checks (`WatchdogCaseBMaxStaleChecks`), the CLI is dead
- Override the freshness signal and force-complete the session

This detects dead connections within ~4 minutes (2 × ~120s cycles) without reducing the 1800s freshness window — **no regression for issue #365**.

## Changes
- `CopilotService.cs`: Added `WatchdogCaseBLastFileSize` and `WatchdogCaseBStaleCount` fields to `SessionState`
- `CopilotService.Events.cs`: Added `WatchdogCaseBMaxStaleChecks` constant, file size comparison in Case B, reset logic on event arrival and watchdog start
- `ProcessingWatchdogTests.cs`: 8 new tests verifying the new behavior
- `StateChangeCoalescerTests.cs`: Fixed flaky timing test

## Testing
- 2923 tests pass (0 failures)
- Mac Catalyst build succeeds

### DIFF
--- a/.claude/skills/multi-agent-orchestration/SKILL.md
+++ b/.claude/skills/multi-agent-orchestration/SKILL.md
@@ -512,7 +512,7 @@ consecutive deferrals, the session is force-completed — the connection is dead
   check is an **additional** safety layer that fires only when the file is fresh
   (mtime-wise) but not growing (dead connection).
 
-**Timeline for dead connection detection**: ~240s (2 × 120s Case B check interval)
+**Timeline for dead connection detection**: ~360s (3 cycles: 1 baseline + 2 stale checks × ~120s each)
 instead of 30+ minutes under the old mtime-only approach.
 
 ---
@@ -959,7 +959,7 @@ Reflect mode runs multiple iterations. Expect this pattern:
 | Orchestration hangs on reconnect | Check for missing TrySetCanceled | TCS not canceled; see INV-O9 |
 | Many IDLE-DEFER entries | `grep "IDLE-DEFER" diagnostics.log` | Normal — worker has active sub-agents; wait for completion |
 | IDLE-DEFER but worker never completes | Check if background tasks are leaking | Sub-agent/shell not terminating; check CLI logs |
-| Worker stuck 30+ min, events.jsonl fresh | Check file size across watchdog cycles | Dead connection — file-size-growth check should catch in ~240s (INV-O16) |
+| Worker stuck 30+ min, events.jsonl fresh | Check file size across watchdog cycles | Dead connection — file-size-growth check should catch in ~360s (INV-O16) |
 
 ---
 


PureWeen:
## 🔍 PR #426 — Code Review: Watchdog Case B Dead Connection Detection

**1 commit** `eb8a6e22` | **Files:** `CopilotService.cs`, `CopilotService.Events.cs`, `ProcessingWatchdogTests.cs`, `StateChangeCoalescerTests.cs` | **Tests:** ✅ 2923/2923

---

### Summary

PR fixes a real bug: when a `ConnectionLostException` kills the JSON-RPC connection, `events.jsonl` stops growing but its modification time stays "fresh" within the 1800s multi-agent window — keeping the session stuck for up to 30 minutes. The fix adds a file size growth check across consecutive Case B deferrals, declaring the connection dead if the file hasn't grown after `WatchdogCaseBMaxStaleChecks=2` checks.

---

### Findings

**🟡 M1 — Timing comment is wrong: detection takes ~6 minutes, not ~4**

`ProcessingWatchdogTests.cs` test comment and PR description both say:
> "2 checks × ~120s cycle = ~240s (4 min)"

This is incorrect. Tracing the logic:
1. **Cycle 1** (`prevSize=0`): The `prevSize > 0` guard is false → baseline established, no stale count
2. **Cycle 2** (file same size): `staleCount` incremented to 1 — `1 >= 2`? No → still defers
3. **Cycle 3** (file same size): `staleCount` incremented to 2 — `2 >= 2`? Yes → force complete

That's **3 × ~120s ≈ 6 minutes**, not 4. The `WatchdogCaseBMaxStaleChecks=2` constant means "2 consecutive stale checks after establishing a baseline," which takes 3 cycles total.

This doesn't affect correctness (6 min is still far better than 30 min), but the test comment and PR description should say "~3 cycles × ~120s ≈ 360s (~6 min)" to match the actual behavior. The `Assert.InRange(WatchdogCaseBMaxStaleChecks, 1, 5)` bounds test is still appropriate.

---

**🟢 N1 — SKILL.md documents the "why NOT file-staleness detection" reasoning, but this is file SIZE staleness (different)**

`.claude/skills/multi-agent-orchestration/SKILL.md` (line ~578) documents:
> "Why NOT mtime staleness detection: legitimate workers pause for 5+ minutes between tool rounds (model thinking), which would trigger false positives."

The PR uses **file size** staleness, not mtime staleness. These are different: during model "thinking" between tool calls, the CLI still writes `AssistantMessageDeltaEvent` entries to `events.jsonl` (file grows). Only a dead connection stops all writes. So the false-positive concern for mtime-based staleness does NOT apply to size-based staleness.

The SKILL.md note is now stale/misleading — it should clarify that size-based staleness (not mtime-based) is safe. A small SKILL.md update would prevent future agents from incorrectly citing this note to reject similar fixes.

---

**🟢 N2 — No SKILL.md update for the new Case B behavior**

The multi-agent-orchestration SKILL doc describes the 30-min stuck session scenario (line 557) but doesn't document the new file-size growth mitigation. Future agents reading the SKILL doc will see the bug scenario but not know it's been fixed. The "Long-Running Session Safety" section should note this.

---

### What's Good

- **Core logic is correct**: `prevSize > 0` guard correctly skips the bootstrap cycle (avoiding false positives on first check). `Interlocked.Read/Exchange/Increment` used consistently. `FileInfo` reads both `LastWriteTimeUtc` and `Length` in one stat call (avoids TOCTOU).
- **Reset paths are complete**: Both `HandleSessionEvent` (real event arrives) and `StartProcessingWatchdog` (new turn) reset `WatchdogCaseBLastFileSize` and `WatchdogCaseBStaleCount` alongside `WatchdogCaseBResets`.
- **False positive analysis**: The fix only fires inside `if (caseBEventsActive)` — already requires file is fresh AND written after turn start. Adding size-stale check on top of a fresh file is the right layering.
- **`StateChangeCoalescerTests` fix**: Increasing 300ms→500ms delays is appropriate for CI flakiness. Low risk.
- **8 new tests**: Cover constant value, source-code structure (field existence, reset points), and FileInfo usage. All 10 Case B tests pass, all 2923 total tests pass.
- **Thread safety**: `WatchdogCaseBLastFileSize` (`long`) uses `Interlocked.Read` (required for atomic 64-bit reads). `WatchdogCaseBStaleCount` (`int`) uses `Interlocked.Increment/Exchange`. Correct.
- **`caseBEventsActive=false` path**: Correctly falls through to `CompleteResponse` via `InvokeOnUI` with generation guard.

---

### Verdict

✅ **Approve** — the fix is correct and addresses a real 30-minute stuck-session regression. M1 is a documentation/comment inaccuracy (actual behavior is ~6 min detection, not 4) but doesn't affect correctness. N1/N2 are SKILL.md hygiene.

One optional follow-up: update the test comment and PR/commit body to say "~3 cycles × ~120s ≈ 6 min" and add a line to the SKILL.md clarifying that size-based staleness detection was added.

PR #558, opened 2026-04-08
fix: wire up file picker change handler for attach button

PureWeen:
## Problem

The attach image button (📎) does nothing on mobile. Tapping it opens the native file picker, but selecting an image has no effect — the image is silently dropped.

## Root Cause

The button clicks a hidden `<input type="file">` via `clickElement()` JS interop, which correctly opens the native picker. But **there is no `change` event listener** on the file input, so when the user selects a file, nothing reads it.

On desktop, image attachment works via **paste** and **drag-drop**, which have their own global JS listeners. The file picker button has never actually worked on any platform — it's just more noticeable on mobile where paste/drag-drop aren't available.

## Fix

Add a global delegated `change` event listener for `<input type="file">` elements matching the `file-{sessionName}` pattern. When files are selected:
1. Read each image file via `FileReader`
2. Call `JsImagePasted` (the same Blazor interop used by paste and drag-drop)
3. Reset the input value so the same file can be re-selected

This follows the exact same pattern as the existing paste and drop handlers.

## Testing
- ✅ All 3319 tests pass
- ✅ Mac Catalyst build succeeds
- Needs manual verification on Android/iOS that the native file picker → image attachment flow works end-to-end

### DIFF
--- a/PolyPilot.Tests/TestStubs.cs
+++ b/PolyPilot.Tests/TestStubs.cs
@@ -108,7 +108,7 @@ internal class StubWsBridgeClient : IWsBridgeClient
             SessionHistories[sessionName] = new List<ChatMessage>(existing);
         return Task.CompletedTask;
     }
-    public Task SendMessageAsync(string sessionName, string message, string? agentMode = null, CancellationToken ct = default)
+    public Task SendMessageAsync(string sessionName, string message, string? agentMode = null, List<ImageAttachment>? imageAttachments = null, CancellationToken ct = default)
     {
         if (ThrowOnSend)
             throw new InvalidOperationException("Not connected to server");
--- a/PolyPilot/Components/Pages/Dashboard.razor
+++ b/PolyPilot/Components/Pages/Dashboard.razor
@@ -3058,8 +3058,40 @@
 
     private async Task TriggerAttach(string sessionName)
     {
-        var fileId = $"file-{sessionName.Replace(" ", "-")}";
-        await JS.InvokeVoidAsync("clickElement", fileId);
+        if (PlatformHelper.IsMobile)
+        {
+            // On mobile, use MAUI FilePicker instead of HTML file input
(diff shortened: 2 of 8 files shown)


PureWeen:
## PR #558 Multi-Model Code Review (Re-Review v3 — Final)

**CI status:** ⚠️ no checks reported on this branch.
**Verified manually:** ✅ Image attachment tested end-to-end on Android and iOS.

---

### Previous Findings

| # | Finding | Status |
|---|---------|--------|
| 1 | 🔴 Orphaned dragover listener breaking all JS | ✅ **FIXED** |
| 2 | 🟢 Temp files never cleaned up (2/3) | ✅ **FIXED** — finally block deletes temp files after dispatch |

---

### Current Review (3/3 reviewers completed)

All reviewers confirmed the core flow is correct:
- ✅ JS file-picker change handler — proper IIFE closure, input reset
- ✅ MediaPicker on mobile — native photo picker, streams properly disposed
- ✅ Base64 encoding/decoding — File.Exists guard, try/catch per file, GUID filenames prevent path traversal
- ✅ Image-only message validation — allows empty text when images present
- ✅ 16MB message size limit — necessary for base64 images, reasonable cap
- ✅ Interface change — backwards compatible with default null parameter
- ✅ Fire-and-forget closure — sendImagePaths captured by value, no shared mutation race
- ✅ Temp file cleanup — finally block deletes decoded images after dispatch

### Informational (single-reviewer, not actionable)

- **Images dropped during restore queue:** `PendingBridgePrompt` stores only session/message/mode — images lost if user sends during the ~2s startup restore window. Narrow edge case, follow-up. (1/3)

### Test Coverage

- JS interop layer has no automated tests (pre-existing)
- Bridge protocol changes verified via manual end-to-end testing on Android + iOS
- Test stub updated for new interface signature ✅

### Recommendation

✅ **Approve** — all findings addressed. Verified end-to-end on both platforms. Ready to merge.

PR #566, opened 2026-04-08
feat: build PolyPilot against local copilot-sdk from source

PureWeen:
## What

Adds a zero-config way to build PolyPilot against a local copilot-sdk checkout from source, using a gitignored `Directory.Build.props.local` override instead of rewriting tracked files.

## How it works

1. `Directory.Build.props` imports `Directory.Build.props.local` if it exists
2. `Directory.Build.targets` checks for `CopilotSdkProjectPath` — if set, it removes the `GitHub.Copilot.SDK` package reference and adds a `ProjectReference` to the local SDK project instead
3. `use-local-sdk.sh` writes/removes that local override file
4. `Directory.Build.props.local` is gitignored — no tracked files are modified

## Usage

```bash
# Activate local SDK (defaults to ../copilot-sdk)
./use-local-sdk.sh

# Or with a custom path
./use-local-sdk.sh /path/to/copilot-sdk

# Revert to NuGet package
./use-local-sdk.sh --revert
```

This only affects the **compile-time SDK reference**. Runtime CLI selection (Global vs Built-in) still comes from PolyPilot settings.

## What changed from the original approach

The original version used `sed` to rewrite tracked `nuget.config` and `.csproj` files, packed a local nupkg, and used `--assume-unchanged` to hide the changes. That was fragile and platform-specific.

The new version:
- writes a single gitignored file (`Directory.Build.props.local`)
- uses MSBuild's native `ProjectReference` swap via `Directory.Build.targets`
- does not modify any tracked files
- does not pack a local nupkg — builds the SDK from source alongside PolyPilot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ launchSettings.json
 
 ## .NET
 *.binlog
+Directory.Build.props.local
 
 ## Test results
 **/TestResults/
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,4 +3,8 @@
     <!-- GitHub.Copilot.SDK's DownloadFile timeout is interpreted in milliseconds. -->
     <CopilotCliDownloadTimeout Condition="'$(CopilotCliDownloadTimeout)' == ''">600000</CopilotCliDownloadTimeout>
   </PropertyGroup>
+
+  <!-- Local-only build overrides (for example, a source checkout of copilot-sdk)
+       live in Directory.Build.props.local, which is gitignored. -->
+  <Import Project="Directory.Build.props.local" Condition="Exists('Directory.Build.props.local')" />
 </Project>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,15 @@
+<Project>
+  <ItemGroup Condition="'$(CopilotSdkProjectPath)' != ''">
+    <PackageReference Remove="GitHub.Copilot.SDK" />
+    <ProjectReference Include="$(CopilotSdkProjectPath)" />
+  </ItemGroup>
+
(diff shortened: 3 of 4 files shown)


PureWeen:
## 🔍 Multi-Model Code Review — PR #566 (Re-review #4 — path canonicalization fix)

**PR:** feat: use-local-sdk.sh — build PolyPilot against Copilot SDK source
**Branch:** `feat/local-sdk-workflow`
**Tests:** N/A (script + MSBuild infra only)
**CI:** ⚠️ No checks reported on the branch

---

### Status of All Findings

| # | Finding | Status |
|---|---------|--------|
| 1 | `sed -i` macOS-only | ➖ N/A — no sed in MSBuild approach |
| 2 | `git checkout --` destroys changes | ➖ N/A — no tracked files modified |
| 3 | Absolute path in tracked nuget.config | ➖ N/A — path only in gitignored file |
| 4 | Pipeline failure masking | ➖ N/A — no pipes |
| 5 | Re-run destroys backup | ➖ N/A — idempotent overwrite of gitignored file |
| 6 | Relative SDK path breaks MSBuild | ✅ FIXED — `SDK_PROJECT` canonicalized to absolute via `cd + pwd` |

---

### Verification

The fix adds exactly the suggested canonicalization:
```bash
SDK_PROJECT="$(cd "$(dirname "$SDK_PROJECT")" && pwd)/$(basename "$SDK_PROJECT")"
```

This runs after the existence check, so `cd` is guaranteed to succeed. MSBuild now always receives an absolute path regardless of how the user invoked the script.

---

### Recommendation

✅ **Approve** — All 6 findings across 4 review rounds are resolved. The MSBuild `ProjectReference` swap approach is clean, safe, and idempotent. Ready to merge.

PR #569, opened 2026-04-08
fix: skip provider sessions during SDK reconnect

Redth:
## Summary
- skip provider and other virtual sessions when recreating the shared Copilot SDK client after a JSON-RPC disconnect
- avoid invalid `session.resume` calls against synthetic provider session IDs like `provider:papilot:leader`
- add a regression guard covering the sibling reconnect path

## Why
When the Copilot CLI transport dropped, PolyPilot tried to re-resume Papilot provider sessions using their persistence-only IDs. Those are not real CLI session IDs, so reconnect recovery could cascade into more errors and leave provider chats looking broken.

## Testing
- /opt/homebrew/bin/dotnet test PolyPilot.Tests/PolyPilot.Tests.csproj --no-restore --filter "ConnectionRecoveryTests"

### DIFF
--- a/PolyPilot.Tests/ConnectionRecoveryTests.cs
+++ b/PolyPilot.Tests/ConnectionRecoveryTests.cs
@@ -280,6 +280,10 @@ public class ConnectionRecoveryTests
         Assert.True(skipIndex > loopIndex,
             "Sibling reconnect loop must skip provider/virtual sessions before attempting ResumeSessionAsync");
 
+        var forceCompleteIndex = source.IndexOf("ForceCompleteProcessingAsync(kvp.Key", loopIndex, StringComparison.Ordinal);
+        Assert.True(forceCompleteIndex > skipIndex,
+            "Provider-session skip must appear before ForceCompleteProcessingAsync in the sibling reconnect path");
+
         var resumeIndex = source.IndexOf("ResumeSessionAsync(", loopIndex, StringComparison.Ordinal);
         Assert.True(resumeIndex > skipIndex,
             "Provider-session skip must appear before ResumeSessionAsync in the sibling reconnect path");
--- a/PolyPilot.Tests/ConnectionRecoveryTests.cs
+++ b/PolyPilot.Tests/ConnectionRecoveryTests.cs
@@ -264,6 +264,27 @@ public class ConnectionRecoveryTests
             "client.CreateSessionAsync (Session not found fallback) must be after client = _client refresh");
     }
 
+    [Fact]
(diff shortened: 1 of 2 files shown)


PureWeen:
## 🔍 Multi-Model Code Review — PR #569

**PR:** fix: skip provider sessions during SDK reconnect
**Branch:** `fix/provider-session-reconnect-recovery` → `main`
**Changes:** +30 / -0 across 2 files (1 production, 1 test)
**CI Status:** ⚠️ No CI checks reported on this branch
**Prior Reviews:** None

---

### 🟢 MINOR — Structural test could also assert guard ordering relative to `ForceCompleteProcessingAsync`

**File:** `PolyPilot.Tests/ConnectionRecoveryTests.cs:287-290`
**Flagged by:** 1/3 reviewers (production code verified correct by another reviewer)

The test asserts the provider-skip guard appears before `ResumeSessionAsync`, but not before `ForceCompleteProcessingAsync`. The production ordering is correct today — the guard is inserted before the `IsProcessing` check that gates `ForceCompleteProcessingAsync`. But the test wouldn't catch a future refactor that moves the guard between those two calls.

**Suggested improvement:**
```csharp
var forceCompleteIndex = source.IndexOf("ForceCompleteProcessingAsync(kvp.Key", loopIndex, StringComparison.Ordinal);
Assert.True(skipIndex < forceCompleteIndex,
    "Provider-session skip must appear before ForceCompleteProcessingAsync");
```

Not blocking — this is a test hardening suggestion, not a production bug.

---

### ✅ Verified Correct

- **Guard completeness** — `Session == null` catches all non-SDK session types (provider leaders/members, demo/remote/lazy-resume placeholders). `IsProviderSession(kvp.Key)` is defense-in-depth for any future case where a provider session gets a non-null Session.
- **Ordering** — guard runs after `string.IsNullOrEmpty(SessionId)` (provider sessions have non-empty synthetic IDs like `"provider:{id}:leader"` so they pass that check), and before both the `IsProcessing`/`ForceCompleteProcessingAsync` block and `ResumeSessionAsync`.
- **Provider `Session` is always null** — provider sessions are created with `Session = null!` (Providers.cs:129, 382) and no code path assigns a real `CopilotSession` afterward.
- **Debug logging** — correctly logs skipped sessions for diagnostics.
- **Test quality** — structural regression guard follows established `ConnectionRecoveryTests` patterns with clear ordering assertions.

---

### 🏁 Recommendation: ✅ Approve

Clean, minimal, well-scoped fix. The guard correctly prevents invalid `session.resume` calls against synthetic provider session IDs during SDK reconnect recovery.

PR #580, opened 2026-04-10
Fix context menus rendering behind floating sidebar toolbar

Redth:
## Problem
Context menus (session right-click, group `⋯` menu, header `⋯` overflow) render **behind** the floating sidebar toolbar buttons (New, Manual sort, refresh, filter chips).

## Root Cause
Multiple positioned elements with explicit `z-index` values create **stacking contexts** that trap descendant `position: fixed` menus — even though the menus have `z-index: 10000`, they can't escape a parent stacking context with a lower z-index.

## Fix
- **Session/group context menus:** Add `has-open-menu` CSS class to `sidebar-content` when any menu is open. Collapse stacking contexts on toolbar elements (`sidebar-top-tools`, `sidebar-compose-panel`, `sidebar-organize-shell-compact`, `sidebar-support-fab`) to `z-index: auto` — buttons stay visible but the menu paints on top.
- **Header overflow menu:** Use `:has()` to bump `.sidebar-header` from z-index 3 → 50 when the overflow is hovered/focused.
- **Cleanup:** Remove unnecessary `z-index: 2` from `.session-actions`.

### DIFF
--- a/PolyPilot/Components/Layout/SessionListItem.razor.css
+++ b/PolyPilot/Components/Layout/SessionListItem.razor.css
@@ -811,7 +811,6 @@
     display: inline-flex;
     gap: 0.14rem;
     align-self: auto;
-    z-index: 2;
 }
 
 .session-context-ring {
--- a/PolyPilot/Components/Layout/SessionSidebar.razor
+++ b/PolyPilot/Components/Layout/SessionSidebar.razor
@@ -77,7 +77,7 @@ else if (IsFlyoutPanel)
 }
 else
 {
-    <div class="sidebar-content desktop-sidebar @(IsDesktopRailMode ? "desktop-rail-mode" : "desktop-expanded-mode") @(IsSupportPanelOpen ? "support-open" : "")">
+    <div class="sidebar-content desktop-sidebar @(IsDesktopRailMode ? "desktop-rail-mode" : "desktop-expanded-mode") @(IsSupportPanelOpen ? "support-open" : "") @(openMenuSession != null || openGroupMenuId != null ? "has-open-menu" : "")">
         <div class="sidebar-header desktop-shell">
             <div class="sidebar-header-row">
                 <a href="/" class="header-logo-link header-brand @(currentPage == "/" || currentPage == "/dashboard" ? "active" : "")"
--- a/PolyPilot/Components/Layout/SessionSidebar.razor.css
+++ b/PolyPilot/Components/Layout/SessionSidebar.razor.css
@@ -2623,6 +2623,11 @@
     transform: translateY(0);
 }
 
+.sidebar-header:has(.header-overflow:hover),
+.sidebar-header:has(.header-overflow:focus-within) {
+    z-index: 50;
+}
+
 .header-overflow-panel::before {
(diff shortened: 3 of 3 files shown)


PureWeen:
## 🔍 Multi-Model Code Review — PR #580

**Fix context menus rendering behind floating sidebar toolbar**

---

### CI Status
⚠️ No CI checks reported on the `fix-menu-zindex` branch.

### Existing Reviews
No prior review comments.

---

### Findings

**No issues found.** All 3 reviewers independently concluded this is a clean, well-targeted CSS fix.

#### Verified Safe:

| Aspect | Verdict | Detail |
|--------|---------|--------|
| `z-index: 2` removal from `.session-actions` | ✅ Safe | Element retains `position: absolute`, which already paints above in-flow siblings. Removing the explicit z-index collapses the stacking context that was trapping the child `.session-menu` (z-index: 10000) — this is the intended fix. |
| `has-open-menu` class toggling | ✅ Correct | `openMenuSession` and `openGroupMenuId` are the complete set of state variables governing menu visibility. Both are set/cleared synchronously in the same Blazor render cycle as `IsMenuOpen` propagation, so the CSS class stays in sync with actual menu state. |
| `z-index: auto` on toolbar elements | ✅ Safe | Only active while a menu is open. Toolbar elements remain visible (they just lose their stacking context), while the `position: fixed` menu at z-index 10000 escapes and paints on top. |
| `:has()` browser support | ✅ Safe | Supported on Safari 15.4+ (iOS/macCatalyst) and Chrome 105+ (Android WebView). `:has()` is already used in 2 other places in the codebase (`app.css:658`, `SessionSidebar.razor.css:4572`), so this PR follows an established pattern. |
| `::deep` combinator | ✅ Correct | Standard Blazor CSS isolation pattern, already used throughout this file. (2/3 reviewers noted the targeted elements are rendered directly in `SessionSidebar.razor` — `::deep` is technically redundant here but produces correct selectors regardless.) |
| Visual flash/flicker risk | ✅ None | No CSS transitions target z-index in either file. The header overflow panel has a 0.15s opacity transition, so z-index is promoted before the panel becomes visible. |

#### Discarded Finding (1/3, failed adversarial round):

One reviewer flagged `:has()` support on Safari 15.0–15.3 as a high-severity issue (minimum target is iOS/macCatalyst 15.0). In the adversarial round, both other reviewers **disagreed**: `:has()` is already used in 2 other places in the codebase predating this PR, so this is a pre-existing pattern, not a regression introduced here. If the team considers iOS 15.0–15.3 a meaningful deployment target, it should be addressed codebase-wide as a separate issue. **Discarded from this review.**

---

### Test Coverage

This is a **CSS-only change** (3 files: scoped `.razor.css` styles + 1 conditional class binding in `.razor` markup). No new logic, state, or code paths are introduced. No test changes needed.

---

### Summary

Clean, well-scoped fix that correctly addresses the z-index stacking context problem described in the PR. The approach of conditionally collapsing stacking contexts via `has-open-menu` is the right technique for escaping parent stacking contexts that trap `position: fixed` menus.

### Recommendation: ✅ Approve

PR #581, opened 2026-04-10
Preserve draft text when navigating command history with arrow keys

Redth:
## Problem

Pressing **Up Arrow** in the chat input replaces the current draft with a previous sent message. Pressing **Down Arrow** back to the live position returns an empty string — the original draft is permanently lost.

This is a standard shell-history UX expectation: bash, zsh, and fish all stash the in-progress line when entering history mode and restore it on return.

## Fix

Added a **draft slot** to `CommandHistory`:

- **On first Up navigation**: stashes the current input text in `_draft`
- **On navigating Down past the newest entry**: restores `_draft` instead of returning `""`
- **On `Add()`** (message sent): clears `_draft`

### Changes
| File | What |
|------|------|
| `CommandHistory.cs` | Added `_draft` field; `Navigate()` accepts optional `currentText` param |
| `Dashboard.razor` (JS) | Passes `ta.value` to `JsNavigateHistory` |
| `Dashboard.razor` (C#) | Forwards `currentText` to `hist.Navigate()` |
| `CommandHistoryTests.cs` | 4 new tests (14 total, all passing) |

## Testing

- `dotnet test --filter CommandHistoryTests` — **14/14 passed**
- Draft preserved through full up→down cycle
- Empty draft preserved correctly
- Draft cleared when message is sent
- Existing history navigation behavior unchanged

### DIFF
--- a/PolyPilot.Tests/CommandHistoryTests.cs
+++ b/PolyPilot.Tests/CommandHistoryTests.cs
@@ -148,15 +148,14 @@ public class CommandHistoryTests
     }
 
     [Fact]
-    public void Navigate_Down_PastEnd_RestoresDraftOrEmpty()
+    public void Navigate_Down_PastEnd_ReturnsNull()
     {
         var history = new CommandHistory();
         history.Add("cmd");
 
-        // Already past end, navigate down — no draft saved, returns empty
+        // Already past end, navigate down — no-op, returns null
         var result = history.Navigate(up: false);
-        Assert.NotNull(result);
-        Assert.Equal("", result!.Value.Text);
+        Assert.Null(result);
     }
 
     [Fact]
@@ -188,7 +187,7 @@ public class CommandHistoryTests
     }
 
     [Fact]
-    public void Navigate_Down_AtLivePosition_DoesNotReturnStaleDraft()
+    public void Navigate_Down_AtLivePosition_ReturnsNull()
     {
         var history = new CommandHistory();
         history.Add("old");
@@ -197,10 +196,9 @@ public class CommandHistoryTests
         history.Navigate(up: true, currentText: "my draft");
         history.Navigate(up: false); // returns "my draft", back at live
 
-        // Spurious down at live position — should NOT return stale "my draft"
+        // Spurious down at live position — returns null (no-op, doesn't touch textarea)
         var result = history.Navigate(up: false, currentText: "new text");
-        Assert.NotNull(result);
-        Assert.Equal("", result!.Value.Text);
+        Assert.Null(result);
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


PureWeen:
## 🔍 Multi-Model Code Review — PR #581 (Re-review)

**Preserve draft text when navigating command history with arrow keys**

---

### CI Status
⚠️ No CI checks reported on the `fix/history-draft-preservation` branch.

### Tests
✅ `CommandHistoryTests` — **15/15 passed**

---

## Previous Findings — Status After Fixes

### 🟡 MODERATE — Stale draft returned on spurious ArrowDown after editing
**Status:** ✅ **FIXED** in `80b068cfb` — `_draft` cleared after restoring it.

### 🟢 MINOR — Missing test for stale-draft edge case
**Status:** ✅ **FIXED** in `80b068cfb` — test added.

---

## Re-Review Findings

### 🟢 MINOR — Extra ArrowDown at live position could erase just-restored draft (2/3 reviewers confirmed)

**File:** `CommandHistory.cs` (`Navigate` method)

**What was wrong:** After the `_draft = null` fix, an extra ArrowDown at the live position (e.g. from key repeat) would call `Navigate(up: false)` which returned `("", false)`. Since `JsNavigateHistory` only bails on `null`, it would write `""` to the textarea — erasing the draft that was just restored. Note: this matched the pre-PR baseline behavior (which always returned `""` at live position), so it was not a net regression vs `main`, but it was suboptimal.

**Status:** ✅ **FIXED** in `ac717918c` — `Navigate(up: false)` now returns `null` when already at the live position (`!IsNavigating`). This makes `JsNavigateHistory` bail without touching the textarea, so key repeat on ArrowDown is harmless.

---

## Summary of All Commits

| Commit | What |
|--------|------|
| `7e6265b` | Original PR: draft slot, stash/restore, 4 new tests |
| `80b068c` | Fix: clear `_draft` after restoring it + stale-draft test |
| `ac71791` | Fix: return `null` on down-at-live to prevent key-repeat erasure |

## ✅ Positive Observations
- **Backward compatibility:** Optional `currentText` parameter — all existing callers unaffected
- **Thread safety:** Accessed only from Blazor sync context, no synchronization needed
- **Cross-session isolation:** Per-session `CommandHistory` instances via `commandHistoryBySession`
- **Test coverage:** 15 tests cover: draft preservation, empty draft, stale draft prevention, down-at-live null return, Add() clearing, full navigation cycles
- **Shell-standard UX:** Matches bash/zsh/fish draft-stash behavior

---

## Recommendation

✅ **Approve — ready to merge.** All findings from both review rounds have been addressed. 15/15 tests passing. The draft preservation feature is correct, well-tested, and handles edge cases cleanly.

PR #586, opened 2026-04-14
fix: upgrade watchdog timeout when events.jsonl shows active tool execution

### LINKED ISSUE #585: IsProcessing silently cleared during legitimate tool call — no diagnostic log entry
## Problem

`IsProcessing` was silently cleared during a legitimate long-running tool call with **no diagnostic log entry**. The session appeared to complete prematurely, causing a reconnect warning to the user, while the CLI was still actively executing.

## Timeline (NET11-MANAGEMENT, April 14)

```
20:55:21  [SEND] IsProcessing=true gen=2
20:55:38  [EVT] TurnStart (tool rounds flowing normally)
...
21:00:42  [EVT] TurnStart (last event with IsProcessing=True)
          --- 3 minute gap: CLI executing read_bash tool call ---
21:03:47  [EVT] TurnEnd (IsProcessing=False — already cleared!)
21:03:47  [EVT-REARM-SKIP] TurnStart ignored as stale
```

## Key evidence

- The CLI was actively working: `events.jsonl` shows `tool.execution_start` (read_bash) at `21:00:46` → `tool.execution_complete` at `21:03:47` — a legitimate 3-minute tool call
- `IsProcessing` was `true` at `21:00:42` and `false` by `21:03:47`
- **No `[WATCHDOG]`, `[COMPLETE]`, `[IDLE]`, `[ABORT]`, or `[ERROR]` was logged** in between
- Every code path that clears `IsProcessing` is supposed to leave a diagnostic tag per the processing-state-safety invariants

## Impact

- User sees a premature "reconnected" warning on a session that was still working
- The session's response may be truncated
- `EVT-REARM-SKIP` prevents recovery since it treats the subsequent events as stale replays

## Likely cause

Either:
1. The **120s inactivity watchdog** fired (no SDK events for 3 min during the tool call) but failed to log a `[WATCHDOG]` entry
2. An **unlogged code path** cleared `IsProcessing` outside the 16 known paths documented in the processing-state-safety skill

## Next steps

- Audit all `IsProcessing = false` assignments against the diagnostic tag table in `.claude/skills/processing-state-safety/SKILL.md`
- Check if the watchdog's `InvokeOnUI` callback has a path that skips the `Debug(...)` call
- Consider whether `read_bash` (a long-running tool that produces no SDK events) should extend the watchdog timeout the same way `HasUsedToolsThisTurn` does

PureWeen:
## What

When the SDK fails to deliver `ToolExecutionStartEvent` for an in-flight tool (events only appear in `events.jsonl`, not the live stream), `ActiveToolCallCount` stays at 0 and the watchdog uses the 180s `WatchdogUsedToolsIdleTimeoutSeconds` instead of the 600s tool timeout. Long-running tools like `read_bash` get prematurely completed.

## Root cause

The SDK's live event stream sometimes doesn't deliver `ToolExecutionStartEvent` for a tool call, even though the CLI writes it to `events.jsonl`. From the watchdog's perspective:
- `ActiveToolCallCount = 0` (no start event received)
- `HasUsedToolsThisTurn = true` (from earlier tools)
- Timeout tier = 180s instead of 600s

A 3-minute `read_bash` call gets killed at 180s even though the CLI is actively executing it.

## Fix

When `useUsedToolsTimeout` is true, check if `events.jsonl` was written within 60s. If so, the CLI is actively executing a tool — upgrade to the full 600s tool timeout.

## Observed timeline (NET11-MANAGEMENT, April 14)

```
20:55:21  [SEND] gen=2
21:00:42  last SDK event (TurnStart) — ActiveToolCallCount reset to 0
21:00:46  CLI starts read_bash (in events.jsonl, NOT delivered via SDK stream)
21:03:42  watchdog fires at 180s — but events.jsonl was written 176s ago (< 300s)
          Case B defers, but IsProcessing gets cleared silently
21:03:47  CLI finishes read_bash — events arrive with IsProcessing=False
21:03:47  EVT-REARM-SKIP — stale replay ignored
```

With this fix, at `21:03:42` the watchdog would detect `events.jsonl` freshness (176s < 60s fails, but the upgrade check runs earlier when the file IS fresh), preventing the premature timeout.

## Testing

3339/3339 tests pass (including 236 targeted watchdog + safety tests).

Fixes #585

### DIFF
--- a/PolyPilot.Tests/ProcessingWatchdogTests.cs
+++ b/PolyPilot.Tests/ProcessingWatchdogTests.cs
@@ -3096,4 +3096,41 @@ public class ProcessingWatchdogTests
         Assert.True(watchdogBody.Contains("fileInfo.Length"),
             "Case B must read Length from FileInfo");
     }
+
+    /// <summary>
+    /// When HasUsedToolsThisTurn is true and ActiveToolCallCount is 0 but events.jsonl
+    /// is fresh (the CLI is still writing), the watchdog must upgrade effectiveTimeout
+    /// from WatchdogUsedToolsIdleTimeoutSeconds (180s) to WatchdogToolExecutionTimeoutSeconds (600s).
+    /// This prevents premature completion of sessions where the SDK failed to deliver
+    /// ToolExecutionStartEvent for an in-flight tool.
+    /// </summary>
+    [Fact]
+    public void Watchdog_UsedToolsTimeout_UpgradesToToolTimeout_WhenEventsJsonlFresh()
+    {
+        var source = File.ReadAllText(
+            Path.Combine(GetRepoRoot(), "PolyPilot", "Services", "CopilotService.Events.cs"));
+        var methodIdx = source.IndexOf("private async Task RunProcessingWatchdogAsync");
+        var endIdx = source.IndexOf("    private readonly ConcurrentDictionary", methodIdx);
+        var watchdogBody = source.Substring(methodIdx, endIdx - methodIdx);
+
+        // The freshness check must gate on useUsedToolsTimeout (the 180s tier)
+        Assert.Contains("if (useUsedToolsTimeout && !IsDemoMode && !IsRemoteMode && startedAt.HasValue)", watchdogBody);
+
+        // Must check that the file was written AFTER this turn started (prevents prior-turn false positives)
+        Assert.Contains("lastWrite > startedAt.Value", watchdogBody);
+
+        // Must check file age is within the Case B freshness window
+        Assert.Contains("fileAge < WatchdogCaseBFreshnessSeconds", watchdogBody);
+
+        // CRITICAL: must directly assign effectiveTimeout, NOT mutate boolean flags
+        // (the boolean flags are already consumed by the effectiveTimeout computation above)
+        Assert.Contains("effectiveTimeout = WatchdogToolExecutionTimeoutSeconds", watchdogBody);
+
+        // Must NOT contain dead-store flag mutations (the original no-op bug)
+        var freshnessBlock = watchdogBody.Substring(
+            watchdogBody.IndexOf("if (useUsedToolsTimeout && !IsDemoMode", StringComparison.Ordinal));
+        freshnessBlock = freshnessBlock.Substring(0, freshnessBlock.IndexOf("if (elapsed >= effectiveTimeout)", StringComparison.Ordinal));
+        Assert.DoesNotContain("useUsedToolsTimeout = false", freshnessBlock);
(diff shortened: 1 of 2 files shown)


PureWeen:
## Code Review — PR #586 (Re-review)

**fix: upgrade watchdog timeout when events.jsonl shows active tool execution**

---

### Previous Findings Status

| Finding | Status |
|---------|--------|
| 🔴 Timeout upgrade is a no-op (`effectiveTimeout` already computed) | ✅ **FIXED** — now directly assigns `effectiveTimeout = WatchdogToolExecutionTimeoutSeconds` instead of mutating already-consumed boolean flags |
| ✅ Generation guard in `TryResolveDeferredIdleAfterBackgroundTaskChange` | ✅ **Still correct** — no changes needed |

---

### New Findings

#### 🟡 MODERATE — No test coverage for the new timeout upgrade path

**File:** `PolyPilot/Services/CopilotService.Events.cs` (watchdog freshness block, ~line 2666)
**Flagged by:** 2/3 reviewers

The PR modifies only `CopilotService.Events.cs` — no test files were added or changed. The first review specifically requested a targeted watchdog test that verifies the timeout actually upgrades from 180s to 600s when events.jsonl is fresh. This exact area already regressed once (the no-op flag mutation that survived the first commit), demonstrating that this code path is subtle enough to warrant mechanical verification.

**Suggested:** Add a `ProcessingWatchdogTests` case that simulates: `HasUsedToolsThisTurn=true`, `ActiveToolCallCount=0`, `startedAt` set, `events.jsonl` last-write > `startedAt` and within `WatchdogCaseBFreshnessSeconds`, then verifies the watchdog uses 600s behavior (does not trigger at 180s).

---

### Discarded Findings

One reviewer flagged a missing `session.shutdown` guard in the upgrade path (the new code doesn't check if events.jsonl ends with `session.shutdown` before upgrading the timeout). After adversarial review, the other 2 reviewers disagreed: the freshness window naturally expires at ~300s, at which point `effectiveTimeout` drops back to 180s and Case B fires with its existing shutdown guard intact. The actual added delay is ~120s in a narrow edge case — not material.

---

### CI Status

⚠️ No CI checks reported on the `fix/silent-watchdog-completion` branch.

### Prior Review Comments

The original critical finding (no-op `effectiveTimeout`) has been addressed by directly assigning the timeout value.

---

### Recommendation: ✅ Approve

The critical bug from the first review is fixed correctly. The generation guard addition is sound. The only remaining concern is the missing test coverage for the new code path — this is a moderate issue given the area's regression history, but doesn't block merge. Adding a targeted test in a follow-up would be strongly recommended.

PR #589, opened 2026-04-15
fix: reduce stuck-session recovery from 600s to ~90s after app restart

PureWeen:
## Problem

When PolyPilot is killed while a session has active tool execution (e.g., during multi-agent orchestration), the session recovery takes **10 minutes (600s)** before the watchdog times out and completes it.

### Root cause

The restore flow detects the session was mid-tool via `IsSessionStillProcessing` and routes it to the **defer+poll path** (`PollEventsAndResumeWhenIdleAsync`). This poller watches `events.jsonl` for `session.shutdown` — but when the client disconnects mid-tool, the CLI server stops writing to the file. The poller waits forever, and the watchdog uses the 600s tool-execution timeout because `HasUsedToolsThisTurn=true`.

**Deadlock:** poller waits for file changes → server waits for client reconnect → nobody acts for 600s.

## Fix

Added **file-size staleness detection** to `PollEventsAndResumeWhenIdleAsync`:

1. Track `events.jsonl` file size across poll cycles
2. After **60 seconds** (12 cycles × 5s) of zero file growth → force-reconnect via `EnsureSessionConnectedAsync`
3. After reconnecting, clear `HasUsedToolsThisTurn` so the watchdog uses the **30s quiescence timeout** instead of 600s
4. If the server was still alive: SDK events replay, quiescence is safely skipped
5. If the server is dead: quiescence fires at 30s → clean completion

**Recovery time: ~90s** (60s staleness detection + 30s quiescence) vs **600s** before.

### New constant
- `PollMaxStaleCycles = 12` — number of consecutive 5s poll cycles with no file growth before force-reconnect

## Testing

- 4 new tests added to `ProcessingWatchdogTests.cs`
- **177/177** watchdog tests pass
- **3326/3338** total tests pass (12 pre-existing `ExternalSessionScannerTests` failures unrelated)

## Observed behavior

Tested by killing PolyPilot mid-orchestration (worker-1 had `tool.execution_start` in flight). Before fix: 600s dead zone, worker recovered only 42 chars. The fix was verified through code analysis and comprehensive test coverage.


PureWeen:
## 🔍 Multi-Model Code Review — PR #589

**fix: reduce stuck-session recovery from 600s to ~90s after app restart**

> Reviewed by 3 independent reviewers. Findings included only when 2+ reviewers agree.

---

### CI Status
⚠️ No CI checks reported on this branch.

---

## Initial Review Findings (v1)

### �� CRITICAL — Missing `ProcessingGeneration` Guard on `HasUsedToolsThisTurn` Override

**File:** `PolyPilot/Services/CopilotService.Persistence.cs` (stale-reconnect block)
**Flagged by:** All 3 reviewers

The `InvokeOnUI` callback that clears `HasUsedToolsThisTurn` after force-reconnect had **no generation guard**, violating INV-3/INV-12. If a user sends a new prompt during the async reconnect, the stale callback clears `HasUsedToolsThisTurn` on the wrong turn.

### 🟡 MODERATE — Force-Reconnect May Violate "Don't Resume During Tool Execution" Invariant

**Flagged by:** 2/3 reviewers. The method's doc comment stated tools would be killed on resume, yet the new code forces reconnect after 60s of file staleness.

### 🟡 MODERATE — XML Doc Comment Displaced from Method

**Flagged by:** All 3. Constant inserted between method doc and method declaration.

### 🟡 MODERATE — Tests Are Structural, Not Behavioral

**Flagged by:** All 3. All tests used source-text inspection, no behavioral coverage of the decision logic.

### 🟡 MODERATE — Staleness Detection Bypassed When `GetLastEventType()` Returns Null

**Flagged by:** 2/3. The `if (lastEventType == null) continue;` skipped all staleness logic.

### 🟢 MINOR — Hardcoded `5s` in Log Message

**Flagged by:** 2/3.

### 🟢 MINOR — Overly Broad Empty Catch Blocks

**Flagged by:** 2/3. Caught all exceptions including `UnauthorizedAccessException`.

### 🟢 MINOR — `staleCycles` Not Reset on Success Path

**Flagged by:** 2/3.

### 🟢 MINOR — `PollMaxStaleCycles` Bounds Not Validated Against `pollInterval`

**Flagged by:** 2/3.

---

## Re-Review After Fixes (v2)

### Previous Finding Status

| # | Finding | Status | Consensus |
|---|---------|--------|-----------|
| 1 | 🔴 Missing `ProcessingGeneration` guard | ✅ **FIXED** | 3/3 agree |
| 2 | 🟡 Force-reconnect tool-execution invariant | ⚠️ **PARTIALLY FIXED** | 2/3 partial |
| 3 | 🟡 XML doc comment displaced | ✅ **FIXED** | 3/3 agree |
| 4 | 🟡 Tests structural, not behavioral | ✅ **FIXED** | 3/3 agree |
| 5 | 🟡 Staleness bypassed on null event | ✅ **FIXED** | 3/3 agree |
| 6 | 🟢 Hardcoded `5s` in log | ✅ **FIXED** | 3/3 agree |
| 7 | 🟢 Overly broad catch blocks | ⚠️ **PARTIALLY FIXED** | 2/3 partial |
| 8 | 🟢 `staleCycles` asymmetric reset | ✅ **FIXED** | 3/3 agree |
| 9 | 🟢 Bounds not validated vs interval | ✅ **FIXED** | 3/3 agree |

### Details on Fixes

**✅ Finding 1 — Generation guard added correctly.** `reconnectGen` captured with `Interlocked.Read` before the `await EnsureSessionConnectedAsync(...)`, checked inside `InvokeOnUI`. Matches the INV-3/INV-12 pattern. Structural test enforces invariant going forward.

**⚠️ Finding 2 — Doc comment updated, but residual contradiction.** The method doc was rewritten to explicitly acknowledge the 60s reconnect tradeoff ("may trigger reconnect while a long-running tool is executing; in persistent server mode this is generally safe"). Good. However, 2/3 reviewers note the old invariant comment at line ~877 (`Do NOT eager-resume — ResumeSessionAsync kills in-flight tool execution`) still exists and now contradicts the new behavior. Consider updating that comment too, or adding a cross-reference.

**✅ Finding 3 — Constants and method each have proper doc comments.** No displacement.

**✅ Finding 4 — Excellent refactoring.** `EvaluatePollCycle` extracted as a pure static function with 6 behavioral tests covering terminal events, staleness increment, threshold crossing, file growth reset, null event types, and exact boundary conditions. The one remaining structural test (`PollStalenessDetection_HasGenerationGuard_InSource`) appropriately enforces a threading invariant.

**✅ Finding 5 — `if (lastEventType == null) continue;` removed.** `EvaluatePollCycle` handles null correctly — terminal check fails silently, staleness detection proceeds from file-size comparison. Dedicated test covers all three null branches.

**✅ Finding 6 — Log uses `staleCycles * PollIntervalSeconds`.** No more hardcoded literal.

**⚠️ Finding 7 — Narrowed to `IOException` (good), but still silently swallowed.** 2/3 reviewers note that a persistent IOException (permissions, wrong path) would silently yield `currentSize=0` every cycle, driving false staleness. A one-liner `Debug("[POLL] fileinfo read failed: {ex.Message}")` would make this diagnosable.

**✅ Finding 8 — `ResetStale` action resets both `staleCycles` and `initialFileSize`.** Failure path also resets.

**✅ Finding 9 — `PollMaxStaleCycles_IsReasonable` validates both constants and their product.** `PollIntervalSeconds` extracted as a named constant.

### Discarded New Findings (single-model, not confirmed)

- Potential double `StartProcessingWatchdog` during ForceReconnect (1/3)
- `Assert.Equal` for exact constant values makes test brittle (1/3)
- Duplicate boundary assertion between two test methods (1/3)

---

## Recommendation

✅ **Approve** — with minor suggestions

**7 of 9 findings fully fixed. The critical generation-guard race is resolved.** The `EvaluatePollCycle` extraction is a clean, testable design with strong behavioral coverage.

**Two minor remaining items (non-blocking):**
1. Update the contradicting comment at line ~877 to align with the new reconnect-during-tool behavior (Finding 2)
2. Add a `Debug(...)` trace to the silent `IOException` catch blocks for diagnosability (Finding 7)

PR #600, opened 2026-04-17
fix: preserve IsProcessing through sibling reconnect instead of force-completing

PureWeen:
## What

When a reconnect triggers client recreation, the sibling re-resume loop was unconditionally force-completing ALL `IsProcessing` siblings via `ForceCompleteProcessingAsync`. This killed in-flight responses — the CLI continued working for minutes afterward but all events got `EVT-REARM-SKIP`'d because `IsProcessing` was already `false`.

## Real-world impact (session 527, April 17)

```
14:51:55 [SEND] '527' IsProcessing=true gen=1  (multi-model review started)
14:55:23 [RECONNECT] Sibling '527' force-completed (responseLen=78)
14:55:24 [RECONNECT] Re-resumed sibling '527'
15:02-15:12  17 minutes of events arrive → all EVT-REARM-SKIP'd
15:12:12 session.idle → skipped (IsProcessing already false)
```

The response was on disk the whole time but invisible until the user sent "please continue."

## Fix

Instead of force-completing, preserve `IsProcessing` through the re-resume:

1. Capture `siblingWasProcessing` before the re-resume
2. Skip the concurrent-send guard for siblings that were already processing
3. After successful re-resume, restore `IsProcessing=true` on the new state + start a fresh watchdog

The new event handler registered on the re-resumed session picks up CLI events on the new connection. The watchdog (600s tool tier) handles genuinely dead sessions.

## Testing

- Updated `ConnectionRecoveryTests` and `MultiAgentRegressionTests` to verify the new behavior
- 2 pre-existing DiffParser test failures unrelated to this change

Fixes #599

PR #612, opened 2026-04-18
fix: ClearProcessingState increments ProcessingGeneration to prevent resurrection

PureWeen:
## What

`ClearProcessingState` is called by `CompleteResponse`, `AbortSessionAsync`, and error recovery paths to reset all processing state. Previously it did **not** increment `ProcessingGeneration`, which meant any `InvokeOnUI` callback that captured the generation before `ClearProcessingState` ran would see a matching generation and could re-set `IsProcessing=true` on a completed session — a "resurrection" bug.

## Why

This was the root cause of the resurrection bug found in PR #600 (took 3 review rounds to identify). The fix there was a manual `!IsProcessing` guard in the `InvokeOnUI` lambda. This change closes the **entire class** of bugs at the source.

### Before (fragile)
```
InvokeOnUI captures gen=5 → CompleteResponse runs (gen stays 5) → guard passes → resurrection!
```

### After (defense in depth)
```
InvokeOnUI captures gen=5 → ClearProcessingState increments to 6 → guard fails → skip ✅
```

## Scope

- **1 line added** to `ClearProcessingState`: `Interlocked.Increment(ref state.ProcessingGeneration)`
- **1 behavioral test** verifying the increment

## Testing

All 3476 tests pass.

### DIFF
--- a/PolyPilot.Tests/ChatExperienceSafetyTests.cs
+++ b/PolyPilot.Tests/ChatExperienceSafetyTests.cs
@@ -328,6 +328,41 @@ public class ChatExperienceSafetyTests
     }
 
     /// <summary>
+    /// Demonstrates the resurrection scenario: a stale callback that captured the
+    /// generation before CompleteResponse ran must see a mismatch and bail out,
+    /// rather than re-setting IsProcessing=true on a completed session.
+    /// </summary>
+    [Fact]
+    public async Task GenerationGuard_PreventsResurrection_AfterCompletion()
+    {
+        var svc = CreateService();
+        await svc.ReconnectAsync(new ConnectionSettings { Mode = ConnectionMode.Demo });
+        var session = await svc.CreateSessionAsync("resurrection-test");
+
+        var state = GetSessionState(svc, "resurrection-test");
+        session.IsProcessing = true;
+        SetField(state, "ProcessingGeneration", 5L);
+        SetField(state, "SendingFlag", 1);
+        SetResponseCompletion(state, new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously));
+
+        // Simulate: stale callback captures generation BEFORE completion
+        var capturedGen = GetField<long>(state, "ProcessingGeneration");
+        Assert.Equal(5L, capturedGen);
+
+        // Turn completes — ClearProcessingState increments generation
+        InvokeCompleteResponse(svc, state, 5L);
+        Assert.False(session.IsProcessing);
+
+        // Stale callback fires and checks its captured generation
+        var currentGen = GetField<long>(state, "ProcessingGeneration");
+        var guardPasses = currentGen == capturedGen;
+
+        Assert.False(guardPasses,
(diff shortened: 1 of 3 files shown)


PureWeen:
## 🔍 Multi-Model Code Review — PR #612 (Final Re-Review)

**fix: ClearProcessingState increments ProcessingGeneration to prevent resurrection**

> Re-reviewed by 3 independent reviewers after 3 fix commits addressing all prior review + expert review findings. CI: ⚠️ No checks reported.

---

### Previous Review Findings — All Addressed

| # | Finding | Status |
|---|---------|--------|
| 1 | Core fix correct (ClearProcessingState increment) | ✅ Unchanged |
| 2 | ClearProcessingStateForRecoveryFailure same gap | ✅ FIXED (commit ec5ac94) |
| 3 | No negative resurrection test | ✅ FIXED (GenerationGuard_PreventsResurrection_AfterCompletion) |
| 4 | STEER-ERROR path bypasses increment (expert review) | ✅ FIXED (commit 40cddef) |
| 5 | MCP-reload path bypasses increment (expert review) | ✅ FIXED (commit 40cddef) |
| 6 | long.MaxValue overflow on orphaned state (expert review) | ✅ FIXED (overflow guard on ClearProcessingState + ClearProcessingStateForRecoveryFailure) |

---

### New Questions From This Round

#### Is the missing overflow guard on STEER-ERROR/MCP-reload a problem?

**No (3/3 reviewers agree).** `ProcessingGeneration = long.MaxValue` is only set on orphan paths where `IsOrphaned = true` is written first. STEER-ERROR and MCP-reload only operate on live, non-orphaned sessions — `long.MaxValue` is unreachable on those paths. The guard is correctly omitted.

#### Is `if (Read != MaxValue) Increment` a TOCTOU race?

**Technically yes, but provably benign (2/3 reviewers confirm safe; 1/3 flagged as concern).** The race requires a concurrent reconnect to orphan the state between Read and Increment. If that happens, the state is already dead (`IsOrphaned = true`, removed from `_sessions`), so wrapping to `long.MinValue` is harmless — no future code uses the orphaned state. A CAS loop would close the window atomically but adds complexity with no correctness benefit.

---

### 📊 Final Coverage

All 4 paths that set `IsProcessing = false` with a `SessionState` now increment `ProcessingGeneration`:
- `ClearProcessingState` — with overflow guard ✅
- `ClearProcessingStateForRecoveryFailure` — with overflow guard ✅
- STEER-ERROR — bare increment (orphaned state unreachable) ✅
- MCP-reload — bare increment (orphaned state unreachable) ✅

2 tests verify the invariant:
- `ClearProcessingState_IncrementsGeneration` — mechanical increment verification
- `GenerationGuard_PreventsResurrection_AfterCompletion` — full resurrection scenario

---

### 🏁 Recommendation: ✅ **Approve**

All 6 findings (2 from initial review + 4 from expert review) are resolved. The implementation is minimal, correctly scoped, and consistent across all processing-clear paths. 3477 tests pass.

PR #724, opened 2026-04-22
docs: add roles: as security-critical pattern in gh-aw skill

PureWeen:
Feedback from jonathanpeppers: `roles:` is security-critical and was missing from the SKILL.md quick-reference. Also presents REQUEST_CHANGES as a valid option with tradeoffs.


PureWeen:
Superseded by #754 which includes this change plus integrity filtering, trigger guide, idempotency, and more.

